Issue #253, opened 2014-10-18
QIF export does not work


agalanin:
Steps to reproduce:
1. Open the application.
2.1. Select Menu -> Export transactions.
2.2. Select "SD Card" destination, QIF export type, "export all" and deselect "delete after export".
2.3. Click "Export" to start export.
2.4. Wait for export finish (message "QIF file exported to ..." displayed).
2.5. Check export location
3.1. Repeat 2.1-2.5 for OFX export type

Expected results:
2.5. QIF file exists in the export location.
3.1. OFX file exists in the export location.

Actual results:
2.5. No QIF file found.
3.1. OFX file exists in the export location.

agalanin:
This bug still present in 1.5.4.

codinguser:
Do you have double entry disabled?

agalanin:
Yes, double entry is disabled.

agalanin:
Fixed in 1.5.5.

PR #283, opened 2015-02-18
QIF export for sinlgle split entry

### LINKED ISSUE #279: Some transactions not exported to QIF
When disabling "Activate Double Entry" option in settings, the transactions created thereafter will not be exported to QIF.
They are missing the transfer account (category) on entry but, more importantly, will not show up in QIF file at all.

fefe982:
Fix #279 .

#279 is caused by the current QIF export logic. When there's N splits in a transaction, only N-1 is recorded in QIF export, as the last one is auto-balanced. When there's only one split, it got lost.

Add special handling for single split transaction in QIF export.

This may also be related to #253 . When I tested for this bug, I found that when there's no transaction to be exported (e.g. all transactions have only split), the export file is not created at all.


codinguser:
Thanks. I will also merge this into hotfix/patches so that I can get it into users hands quickly. 

I am planning in the next bigger update (develop) to add the auto-generation of imbalance accounts which will be hidden (but present) when double-entry is disabled. This means in effect there will always be at least two splits per transaction and should simplify logic in some places. 
Then we would have to update the exporter to ignore imbalance accounts when exporting in single-entry mode. But let's cross that bridge when we get there.

PR #286, opened 2015-02-26
Fixed #285: Delete confirmation blocked by passcode screen

### LINKED ISSUE #285: Delete confirmation blocked by passcode screen
From email bug report: 
"I have a passcode setup and it drops me back out to re-enter the passcode after I've exported transactions, then the delete transactions confirmation has disappeared. I've just tried turning off the passcode and the confirmation box is still there.
I would say this is a bug as it ought to work with passcode enabled, I would suggest sensible behaviour would be to not kick you out to re-enter the passcode after exporting transactions."

aint:


codinguser:
Does this fix mean that if someone starts the GnuCash AccountsActivity with an Intent and supplies the appropriate request code, they can bypass the passcode screen completely?

aint:
Yes, it does.

codinguser:
hmm, I think it is better in that case not to include this patch. 
The original problem which the user reported in #285 will go away anyways and I think better to input the passcode one time too often as one time less. 
But thank you for the quick response to this!

crebiz:
NAVER - http://www.naver.com/
--------------------------------------------

crebiz21@naver.com 님께 보내신 메일 <Re: [gnucash-android] Fixed #285: Delete confirmation blocked by passcode screen (#286)> 이 다음과 같은 이유로 전송 실패했습니다.

--------------------------------------------

받는 사람의 메일박스 용량이 가득 찼습니다.  


--------------------------------------------

PR #288, opened 2015-03-03
Pie chart feature

aint:
Some notes about the implementation:
- works on android 2.2 and above
- accounts balances are taken in absolute values
- displays only the first level of nesting non-placeholder accounts. E.g. if we have the following account hierarchy: `Expences/Auto/Gas/Shell V-Power`, then will be displayed only `Auto` as the sum of all sub accounts.

screen shot 2015-03-03 at 7 22 56 pm
screen shot 2015-03-03 at 7 23 30 pm
screen shot 2015-03-03 at 7 54 33 pm
screen shot 2015-03-03 at 7 54 38 pm
screen shot 2015-03-03 at 7 35 37 pm


codinguser:
Thanks Oleksandr. :+1: 
Looking forward to the other charts which you teased a while back :) ;P

aint:
It's nothing. :smile:

PR #292, opened 2015-03-07
Bugfix 291 drawer_list_item.xml uses theme values only available from API level 11 onwards

### LINKED ISSUE #291: drawer_list_item.xml uses theme values only available from API level 11 onwards
The layout file drawer_list_item.xml uses theme attributes only available from API level 11 onwards, although AndroidManifests declares a minSdkVersion of 9.

This causes GnuCash to crash when starting on my HTC Wildfire (API level 10 - Android 2.3.5).

rivaldi8:
Fixes #291


rivaldi8:
I've just removed the problematic attributes as, at least on my tests with the emulator, I haven't seen any difference in the visualization. However, you may know of a better way to solve the problem.

PR #294, opened 2015-03-15
Allow the application to be installed on the external storage.

### LINKED ISSUE #293: Allow the application to be installed on the external storage
Right now it's not possible to move the application to the external storage (SD card). This is a problem for old phones with small internal memory.

It should be possible to move it to the external storage.

rivaldi8:
Fixes #293


codinguser:
Thanks for the patch. 
However, the app (at least in develop branch) uses features (like service, widgets) which when present, it is recommended that it should not be installed on external storage. See here: http://developer.android.com/guide/topics/data/install-location.html

Issue #189, opened 2014-08-31
Import two different books


fefe982:
We I'm playing with import function, I found that when I import two different book into the App, only accounts from the 1st one is shown on the main Accounts view. However, all accounts can be found from the accounts selection drop-down list. Also, a lot of accounts with the same name show up in the drop-down list. It is hard to tell which is which.

I didn't check the code yet, but I guess it has something to do with the two ROOT accounts from the two different book. Only one of the ROOT account (the 1st) is found and used to create the main Accounts view.

This let me think, should we allow importing two different book? If we allow it, how to merge them?

My idea is to forbid such importing, to make things simple and clear. Importing is only allowed when:
1. The accounts tree in the App is empty.
2. The ROOT account in the App has the same GUID with the imported ROOT account

If the condition is not met, all accounts (and transaction) would be cleared first (with user warned) and then import the new book.

codinguser:
So we agree on import being done with the conditions you mentioned in your original post. 
@fefe982 could you implement this?

fefe982:
Sure.

fefe982:
When implementing this, I just thought of a simpler plan.
What about the content of DB is ALWAYS cleared before the import? I cannot quite think of when a merge should be useful, but a merge can cause problems.

For example:
1) User imported from GnuCash
2) User added some transactions in GnuCash android
3) New transactions are exported and imported in GnuCash
4) User re-import from GnuCash (merge will happen here)

This will result in duplicate transactions, as the GUID generated by Gnucash android and Gnucash cash desktop would be different for the transactions created in Gnucash android.

I also want to find a way to put clearing the DB and addition of new transactions in a single DB transaction when importing, to prevent data loss.

codinguser:
I thought of that too, and it makes sense to always clear the database before import. We will of course always make a backup just in case. So if you can put all that in one database transaction, then go for it. 

I also plan to add automatic scheduled exporting before the 1.6.x release (see the Trello board). 

Btw, can you see if the desktop can open our backup (.gnca) files? Last I tried it crashed, but haven't tried again in a while.

fefe982:
I haven't tried using the desktop to open our backups, but I don't think that will work. A lot of things are different. Such as ROOT account, different GUID schemes (cny-xxxxxx for account GUID), extra XML element (exported, and others), etc.

fefe982:
Updated the PR to delete before import. Be sure to WARN the user about the possible data loss, when RESTORING backups and IMPORTING.

PR #304, opened 2015-03-28
Prevent importing different ROOT account

fefe982:
Stop importing (by throw) if:
1) The import XML has no ROOT, and db is not empty.
     This is a backup restore. Everything should be cleared before import.
2) The import XML has multiple ROOT
     There's an error in import XML
3) The import XML has one ROOT, but it is different form the ROOT in the DB, or DB has no ROOT
     What imported is a different book from the one in the DB.

See discussion in #189.

### DIFF
--- a/app/src/main/java/org/gnucash/android/db/DatabaseAdapter.java
+++ b/app/src/main/java/org/gnucash/android/db/DatabaseAdapter.java
@@ -398,4 +398,25 @@ public abstract class DatabaseAdapter {
     public boolean deleteRecord(@NonNull String uid){
         return deleteRecord(getID(uid));
     }
+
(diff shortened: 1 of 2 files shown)


codinguser:
Thanks. 
I'm on a trip and will merge when I get back to the BatCave :)

fefe982:
Now it will not check for conflict anymore. Instead, the database will be cleaned before anything is imported.

The deletion and importing would happen when the XMl is fully parsed and everything to be imported is gathered. The deletion and importing are put into one DB transaction, so if anything went wrong, the DB would not be affected. It is now easy to achieve using nested transactions, as all the adapters are now using the same DB instance.

PR #306, opened 2015-04-02
auto balance transaction

fefe982:
Auto balance transaction as `saveNewTrasaction`.

### DIFF
--- a/app/src/main/java/org/gnucash/android/model/Money.java
+++ b/app/src/main/java/org/gnucash/android/model/Money.java
@@ -466,4 +466,8 @@ public final class Money implements Comparable<Money>{
     public Money absolute() {
         return new Money(mAmount.abs(), mCurrency);
     }
+
+    public boolean isAmountZero() {
+        return mAmount.equals(BigDecimal.ZERO);
+    }
 }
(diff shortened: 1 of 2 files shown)


codinguser:
Thanks for the changes. 
Would it be easier to do the auto-balancing in the split editor? The split editor already computes the imbalance, so it can easily be converted into a split and added to the split list when the editor is closed. 
Also, split amounts are always saved as absolute values. So the negation does not have any effect. Imbalance accounts of the the Account type BANK, which means they have debit normal balance. 
So if the imbalance split is supposed to reduce the balance, then its type needs to be set to CREDIT and vice versa.

fefe982:
getOrCreateImbalanceAccount may create an extra account in the DB. I don't quite to want to create this extra account if it is not really need (when the transaction is saved, not when the user exits the split editor). I think this should only be called when there's splits that need to be added to the DB from the Imbalance account, and guarded by transaction.

As the split type, shouldn't it be handled automatically by the split constructor? I tried the App, but didn't check the code of the constructor.

fefe982:
And currently there is no way to know whether the user used the split editor when editing an single-split transaction. The original splits would be used, which means if the user did not use the split editor, the edit will be lost. However, as Imbalance split is created, the user will be able to edit it if the future.

It may be not a problem in the future, as we are eliminating single-split transactions, so I didn't try to handle the situation. Just let you know about the current status.

codinguser:
Thanks for the clarification. You're right, it's probably better to handle the auto-balancing during saving.

Yes the split constructor does handle the split sign, but it does so currently in a naive way, not taking the account type into consideration. I will fix it.

PR #314, opened 2015-04-14
GnuCash compatible XML

fefe982:
What made it in-compatible is quite unexpected: it comes from a bunch of typos ...

Added commodity section (turns out it is not quite important, as we now only have ISO currencies).
fix: export accounts in hierarchical order (order by full name). It seems to be required by GnuCash.
fix typo : reconcile_state --> reconcile-state (this is what made GnuCash crash)
fix "typo": gnc:commodity --> act:commodity (for currencies in account definition)
fix "typo": type of acc:parent from string to guid

other fix: create ROOT account if it does not already exits when querying ROOT account GUID. Throw when querying un-exist account GUID. 

It now works for imported account tree, with no template transactions. Multi-currency transactions can also be exported and imported into GnuCash, but in an erroneous state. All non ISO currencies become XXX.


codinguser:
very cool :+1: 
Thanks @fefe982

PR #316, opened 2015-04-14
Bugfix 315: The export progress dialog is never dismissed

### LINKED ISSUE #315: The export progress dialog is never dismissed
After exporting transactions the progress dialog remains visible.

Steps to reproduce:

1. Select "Export" from the menu.
2. Leave all options with defaults (share, QIF, options unchecked).
3. Push "Export".
4. GnuCash shows a progress dialog and initiates the export .

Expected results:

The progress dialog is dismissed when the export finishes.

Actual results:

The progress dialog is never dismissed.

rivaldi8:
Fixes #315


codinguser:
This will cause the app to crash at other times when export is not triggered from within the app or by the user. Maybe better would be to check if the context is from an activity *or fragment* before showing the dialog, or dismissing it

rivaldi8:
Sorry, I misunderstood the code. I've updated the PR with the proper fix.

PR #319, opened 2015-04-22
Import fix and improvement

### LINKED ISSUE #318: Importing old backups (with non-balanced splits) fails
Importing old backups which do not have balanced splits fails due to a "FOREIGN key constraint fails" when bulk adding splits. This I think is the the split account uid foreign constraint.
Auto-balancing takes place during the import process (and that includes creating the imbalance account). So db has to be cleared before importing starts. I've made the adjustments in the develop branch, but still the import fails.
@fefe982 could you look into it? I could send you some old files if you don't have any.

fefe982:
1. replace `equalsIgnoreCase` with `equals`.
   XML is case-sensitive. Gnc Desktop does not do this check either.
2. replace if-else with switch
3. create a hashmap for fast UID query for account
4. FIX: import `split:quantity` instead of `split:value`.
   `split:quantity` uses currency of the split's account, not the currency of the transaction. This will make the account balance calculated in GncA the same as the on Desktop
5. If there's not ROOT in XML (old backups), create one.
6. Create non-exist Imbalance account when auto balancing transactions.
    The account is not created in the DB directly, but will be imported into the DB together with all other account.
    Fixes #318, while also specifying correct transaction UID.
7. Delay UID generation
    UID generation is delay until it is queried. A lot of instances will be created when importing, and the generated UID is useless as it will eventually be replaced by the imported UID. With UID generation delayed, UID generation will not be triggered when importing, which saves some time.
8. Delay generating amount and faster money parse
    `Money` and `BigDecimal` operations (such as `abs`, `divide`, `withCurrency` with generate new instances. By generating the necessary data before any instances is generated, a lot less objects would be generated, which will save some time and space.


codinguser:
I will have to revert the changes to when the model UID is generated (32aaefbb6ea1c308c361d9d18f90ab909a6d2ce8). A lot of code when creating transactions, editing splits etc relies on the fact that there is a UID when a model is created. 
Transactions cannot be created at the moment for example. 
If the savings of not generating the UID immediately are very big, then we can consider adding an overloaded constructor for that purpose.

fefe982:
Although the generation of UID is delayed, whenever anyone calls `getUID()`, he would get a legal UID, which is assigned to the transaction, split, etc. If there is no one using the `mUID` member of `BaseModel` directly, it should not break anything.

codinguser:
I understand, but the models themselves use `mUID` directly. 
In a lot of places, for example, a split is created and added to a new transaction without further ado. The transaction assigns it's uid to the split directly. The UID (of the transaction) is null however, since there has been no need for getUID() until now. And then saving to db fails because the split has a null transaction.

All the models are written this way, with the expectation that `mUID` exists once the model exists. 
I was just wondering if the impact during import is that significant, since the UIDs will have to be generated anyway before the data is saved to the database. 

One solution that comes to mind is, I can  make `mUID` private to the `BaseModel`. That way, it can only be touched through calls to `getUID()`. I'll do that and see if it affects anything else.

PR #328, opened 2015-05-09
Report currency and account color in pie chart

aint:
- Moved report currency option from settings to report screen
 - Added option to use account color in the pie chart

screenshot_2015-05-05-20-13-04


codinguser:
Thanks @aint 
Is using the account color the default option?

aint:
No, it isn't.

Issue #329, opened 2015-05-10
Feature: Set Default Transfer Account to sub-accounts


makarica:
I would like to have an ability to set the default transfer account to all sub-accounts of currently edited account.

vovcacik:
@codinguser hi, is the feature of default transfer accounts also available in the desktop gnucash?

codinguser:
@vovcacik I do not know for certain as I am not very involved in dev of the desktop version.
But my guess is no. It is necessary here because it speeds up transaction entry. On the desktop, you can just type the account name and it autocompletes it for you. The Android version lacks that!

PR #330, opened 2015-05-10
Feature #329: Added ability to set default transfer accounts to sub-accounts

makarica:


codinguser:
I think it is better to dynamically lookup the default transfer account of any ancestor accounts and use that when creating new transactions. (kind of how it is done when setting the title color for the account). That way there is no need to manage settings for sub-accounts individually.
So I would prefer not to merge this right now.

makarica:
Thanks for the input. This sounds better.

codinguser:
Thanks for the patch!

PR #331, opened 2015-05-11
Feature #329: Search also through ancestors for default transfer account

makarica:
Using the suggested approach.


codinguser:
:+1: 
Thanks. I will also make some modifications so that it stops immediately a transfer account is found. That way, the immediate parent's setting will be preferred, not an ancestor account.

makarica:
Great, thanks. I believe that I covered that. 
It should traverse from current account to the root account.
The loop should terminate on first sight of possible transfer account (search while defaultTransferAccountID == 0) or gets to the root account without finding anything.

codinguser:
You are right! That should work just fine.

PR #346, opened 2015-06-26
Ukrainian translation

aint:
Regarding graphs. The pie chart is really buggy. I'll fix it as soon as possible.


codinguser:
:+1:

PR #347, opened 2015-07-09
Migrate from ActionBarSherlock to ActionBarCompat

### LINKED ISSUE #345: Migrate from ActionBarSherlock to ActionBarCompat
We should migrate from `ActionBarSherlock` to `ActionBarCompat`. The former was left unmaintained when Google included `ActionBarCompat` in Android's support library.

This change should solve some problems like #344 and provide better integration with other functionalities.

Some blog posts about how to approach the migration:

* [http://www.grokkingandroid.com/migrating-actionbarsherlock-actionbarcompat/](Migrating from ActionBarSherlock to ActionBarCompat)
* [http://antonioleiva.com/actionbarcompat-migrating-actionbarsherlock/](ActionBarCompat (Part 3): Migrating from ActionBarSherlock)

rivaldi8:
Fixes #345 

Although there's still some work left to do, the bulk of the migration would be done with this PR. The only thing that remains is to fix some small visualisation problems due to the use of different themes and styles by the support library.

As I've just started to learn how themes and styles work in Android, I'd rather merge this now and solve the problems in separate issues.


codinguser:
Cool, thanks :+1: 
I have created a new branch 'feature/material' and pushed the commits to it. 
I think we can do the work for integrating material design in that branch and then merge it in one go to develop.

codinguser:
@rivaldi8 You can checkout the 'feature/material' branch and issue further pull requests to it. 
I will also work on updating the tests

rivaldi8:
Nice! I think I'll start fixing some uses of deprecated classes I found while doing the migration.

Sorry about the tests. I totally forgot to run them before sending the pull request. And yesterday, when I tried I couldn't even run them because of some weird error. It seems today the problem is gone and I can run them again.

codinguser:
No problem :)
I already fixed Travis builds, but there are still more breaking tests (as
expected with such a big theme change)

You can always run "gradle build connectedCheck"
(make sure an emulator is running or device is connected)

PR #349, opened 2015-07-13
Migrate to the non-deprecated version of ActionBarDrawerToggle

### LINKED ISSUE #348: Migrate to the non-deprecated version of ActionBarDrawerToggle
ActionBarDrawerToggle from support library v4 is [deprecated](https://developer.android.com/reference/android/support/v4/app/ActionBarDrawerToggle.html). We should use the one from [support library v7](https://developer.android.com/reference/android/support/v7/app/ActionBarDrawerToggle.html) instead.

rivaldi8:
Fixes #348

### DIFF
--- a/app/src/main/java/org/gnucash/android/ui/BaseDrawerActivity.java
+++ b/app/src/main/java/org/gnucash/android/ui/BaseDrawerActivity.java
@@ -19,7 +19,7 @@ import android.app.Activity;
 import android.content.Intent;
 import android.content.res.Configuration;
 import android.os.Bundle;
-import android.support.v4.app.ActionBarDrawerToggle;
+import android.support.v7.app.ActionBarDrawerToggle;
 import android.support.v4.widget.DrawerLayout;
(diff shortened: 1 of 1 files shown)


codinguser:
The home icon is always an arrow now. Can we have a Hamburger menu?

rivaldi8:
It's strange. I've tested it in two devices (Android 2.3.6 and 4.4.2) and in an emulator (5.0.1). In the three cases a hamburger menu is show that becomes an arrow when the drawer is open. The only exception is the settings activity, which doesn't have a drawer toggle.

What devices have you used for testing? Does it happen in all activities with a drawer toggle?

codinguser:
Strange, now I tested it again (Android 5.1, Nexus 5)  and it works. 
Maybe I did something wrong before.
Thanks!

rivaldi8:
Good :) Not sure if it's what has happened, but sometimes, rebuilding the applications fixes these strange problems.

PR #369, opened 2015-08-29
Charts time ranges

aint:
So far implemented support for time ranges. I will push some changes/refactoring and grouping by month/year/all in a few days.


codinguser:
Thanks :+1:

PR #370, opened 2015-08-31
fix export error in qif

fefe982:
QIF read from view, which now does not contain amount but num/denom


codinguser:
I still get error: "split_quantity_denom" column does not exist, when I try to export QIF

PR #385, opened 2015-09-19
preload exchange rate and converted amount

fefe982:


codinguser:
@fefe982 it seems you may be missing an import of MathContext, at least according to Travis.
Could you check it?

codinguser:
Nevemind, I'll just merge and fix it. I have to push some stuff anyway

PR #393, opened 2015-09-29
Update strings.xml

MarkHaanen:


codinguser:
@MarkHaanen There is an issue on line 63: Send to....
The unicode character &#8230; is for an ellipsis. You should probably maintain it.

Could you edit it and re-send the pull request?

codinguser:
Never mind, I've fixed the breaking strings and merged the changes in the code.

MarkHaanen:
Hmm, strange. &hellip; is the HTML named entity for &#8230; which I only 
changed for my own readability after I noticed &amp; was valid. See 
http://www.fileformat.info/info/unicode/char/2026/index.htm. Does 
Android perhaps not support the full set of entity names? If that's the 
case I shall furthermore resort to the numeric code point.

Regards,
Mark

Ngewi Fet schreef op 29-09-15 om 17:32:

PR #395, opened 2015-10-02
Update french translation

Wicowyn:
French translation updated


codinguser:
Thanks for the translation. 
Files in donottranslate.xml do not need the "translatable" attribute.

PR #400, opened 2015-10-05
fix wrong uid in new split

fefe982:
The generated UUID for a new split is in the wrong format, which will result in an erroneous split when the exported XML is opened by GnuCash desktop.


codinguser:
Very subtle, this bug.
Thanks for catching that Yongxin :)

PR #404, opened 2015-10-11
Fixes widget view on Lollipop 5.1.1 with dense sceen grid

alpha-and-omega:
Hello guys.
# Why I made this patch

After upgrading my Samsung Galaxy S2 to Cyanogen mod 12 with Lollipop 5.1.1 my Gnucash widget became ugly!
With dense home screen grid (5x5 and 6x5) it started to use 2 vertical grid cells instead of 1:

# What I did

I created this patch to solve this problem.
It was not enough to decrease widget minHeight  to 40, because widget button does not fit:

So I made corrections to widget button properties to fix that. I also made widget vertical size resize-able, so anyone can return to 2-row widget if he got used to it :)

Result:

PS: Why button is not square ? Android has no easy way to "centerInside" backgroung of buttons, so it is not trivial to make it square. And squeezed button looks pretty good too.

PS2: This patch does not change widget view on standard screens, widget button is as square as before on standard screens.


codinguser:
Thanks for the patch. I haven't gotten around to testing it yet, but will do so. 
Also, its better to issue patches against the develop branch.

alpha-and-omega:
I am using this patch for 23 days.
One obvious drawback is that sometimes I miss "+" button and touch main area of widget instead.

codinguser:
That was one thing I was worried about when I first saw the screenshot. 
And also that we plan to redo the widget in future releases anyway.

codinguser:
I'm going to close this PR for now. 
Widget work is still far out in the future.

PR #411, opened 2015-10-28
Improve the name of the variable who stores the count of app's widget…

luanfonceca:
Hi folks, i just renamed the variable to a more explicit one.


codinguser:
I'm sorry this patch does not improve the code in a meaningful way.

PR #412, opened 2015-10-28
Returning 0 in 'getTransactionMaxSplitNum'

luanfonceca:
There are a unused if condition, now in any case we return 0, if there was a exception or if there is nothing to get from the cursor.


codinguser:
This patch does not improve the code. Sorry

Issue #357, opened 2015-07-25
Export destination is always "SD Card"


agalanin:
1. Open the application
2. Create some transactions
3. Select Menu -> "Export transactions..."
4. Select "Share file..." as export destination and press "Export" 
5. Check export destination

Expected result:

1. Share file menu is displayed

Actual results:

1. "Exported to : SD card" message displayed
2. New file is created under /mnt/sdcard/org.gnucash.android/exports

nodje:
same issue here

codinguser:
I'm not able to replicate this so far. 
What Android version are you running?

agalanin:
On Tue, 18 Aug 2015 00:47:31 -0700
Ngewi Fet <notifications@github.com> wrote:

> I'm not able to replicate this so far. 
> What Android version are you running?

Android 2.3.6

-- 
Alexander Galanin

nodje:
Android 5.0.2
The problem did not occur prior to the two last updates (before 1.5.0?)
I will post a log trace when I can find the time.

floriankisser:
Same here on CyanogenMod 11 (Android 4.4.4) with gnucash from f-droid. Last export that worked for me was on 07 July, first that didn't on 17 July (don't know the gnucash versions, sry).

lennartS:
Same here on Android 4.2.2 (Fairphone OS). Pretty sure it's broken for me since version 1.6.1.

agalanin:
I am unable to export transactions via E-mail in version 2.0.0-beta1 too:
1. select Menu->Export...
2. export to: send to...
3. select "export all transactions" and QIF format
4. click "export"
Expected results:
1. export menu is displayed
Actual result:
1. "exported to: external service" message is displayed, new file is created under /sdcard/org.gnucash.android/exports.

codinguser:
Thanks for the reports @nodje @floriankisser @lennartS @agalanin  
Could you also let me know what your device locale is set to?

lennartS:
Under „Language & input“ English is set as Language. If you mean a different locale setting, please let me know how to find it and I'll happily supply it.

floriankisser:
de_at. Android 5.1.1 works neither.

codinguser:
no, that's the one. Thanks

agalanin:
Problem is reproducible with both Russian and English (US) locales.

codinguser:
Please send me more details about your device, Android version, rooted or
not, etc.

[...]

PR #419, opened 2015-11-04
fix qif export to SDCard

fefe982:


codinguser:
:+1: thanks :)

rivaldi8:
@fefe982, beware this PR reverts the changes that fixed #357. Also, it's not considered good practice to have a method do more than what it's expected from its name (this confused me a lot while reading the code to fix the bug).

I think we should be careful with SD card operations. It seems there's some inconsistency in how different devices handle it's access. That may be the cause of the export issues.

codinguser:
@rivaldi8 so does it mean that sharing multiple files (split into QIF) is what causes the sharing file to fail.
Also, we rely on the Android getExternalStorageDirectory API. The assumption is that this will return us a location where can safely store the files. 

In the future I'm thinking of changing that to use the cache for the temporary internal export before we move the file to the location desired by the user.

codinguser:
@rivaldi8 I've already merged this request. 
I was just thinking, since you identified the cause of the problem with file sharing to 3rd party services, maybe we can fix it without breaking QIF exports. Any ideas?

rivaldi8:
@codinguser The problem with file sharing is that `copyFile()` doesn't copy the file `src` to `dst`, if `src` is a Qif. Instead, it calls `splitQif()`, which splits `src` into multiple files with `dst` as file name suffix.

The general problem, I think, is that `splitQif()` should be only called once as part of the process of exporting to QIF. Then, all export options should deal with multiple files per export.

@fefe982 Could you describe the issue you are fixing with this PR so I can try to reproduce it?

codinguser:
This issue being fixed is that QIF file was not  being split into the
different currencies.  Or at least, if it was, then only the original
unsplit version was given to the user as the result of the export.

@rivaldi8 Does this mean that the problem of not being able to export is
again present on your device after this patch?

If so then maybe we can do it as you suggest and  treat all exports as a
list of files. Which would be 1 for XML and multiple for OFX.
I'll look into it.

rivaldi8:
Yes, it's failing again.

rivaldi8:
@codinguser, are you working on this? If not, I can look into fixing it.

codinguser:
@rivaldi8 please go ahead and look into it.
Won't have time the next few days
Thanks

codinguser:
@rivaldi8 have you been able to look into it yet?

rivaldi8:
Yes, I'm working on it.

rivaldi8:
@codinguser It have taken more time than I thought but I've finally finished with the fix. I just need some more time to review the changes, do some further clean up, and testing. I'll send a pull request tomorrow.

codinguser:
@rivaldi8 great! looking forward to it...

PR #422, opened 2015-11-09
set default currency as the most used currency

fefe982:
Added code to compute the most used currency in transactions, and set it as default.


codinguser:
thanks :+1:

PR #425, opened 2015-11-15
Bugfix 423  Export to external service is broken in 2.0.1

### LINKED ISSUE #423: Export to external service is broken in 2.0.1
This happens on both of my devices i use it on. I just noticed it with GnuCash Version 2.0.1. Version 2.0.0 still did it for me.

Both devices run Android 4.2.2 with a Kernel Version of 3.4.5. One is a ASUS MeMo Pad 7 HD, the other a Samsung S3 Mini.
GnuCash Android version is 2.0.1 on both.

What do i do:

1. Enter any number of transactions.
2. Slide the side menu out
3. Hit the Export Button
4. Select "Sending" as Export Target. Format is qif; all transactions
5. Hitting "Export."

Expected outcome:
The sharing services show up and i choose one. In my case i would use the official owncloud app to put the qif file on my private server.

What does happen:
Nothing. It claims to export. Then the "Export to external Service" message pops up, but the service selection never shows.

What else did i check:
I tried to send files via other apps. The ES Explorer opened the menu without a problem and let me put files on my owncloud as usual. 

As for now, i workaround the issue by exporting to SD card and sending with another application, but to put it frankly: That's a bit tedious.

If you need anything else from me, just let me know.

rivaldi8:
Fixes #423 

I've tested all exports but the ones to Dropbox, as I don't have an account. You may want to test it first before merging.


codinguser:
Thanks @rivaldi8 
I've merged this commit into develop.
I made some changes to the way the XML exporter handles the export to allow for future flexibility in export destination. 
I also modified the exporter to default to using the cache directory for exports, instead of the SD card, so that we less frequently have to deal with permission issues (esp on Android M)

PR #437, opened 2015-11-25
Features/owncloud sync

fmorato:
Added Owncloud-sync capabilities similar to the existing Drive and Dropbox.

    Features:
    * Possible to specify destivation directory
    * Checks for invalid dir characters
    * Account setup dialog provides instant feedback
    * Should behave as existing Drive and Dropbox sync.
    
    Anti-features:
    * Use SharedPreference with MODE_PRIVATE for Credentials storage.
    * (OwncloudDialogFragment) Account Setup dialog not pretty

It's my first time ever pushing code to a project. Please, let me know if there is anything wrong, or if I should push differently. =]

Feedbacks appreciated!


codinguser:
Cool :+1: thanks. I will review it soon.
Of course it would be easier if OwnCloud supported the Storage Access Framework which we plan to support in the next version. But until then, this would be a great alternative.
One thing though, could you add some unit and UX integration tests to the feature?

codinguser:
Merged into develop branch :+1:

Issue #436, opened 2015-11-25
Bug: Sums set to zero when renaming account


bugmenot2:
Applies to all amounts of all children accounts of the account renamed. OUCH!

fefe982:
Confirmed on the lasted release v2.0.3.

I also found the transaction disappeared.

codinguser:
Fixed in v2.0.4

bugmenot2:
please push this to f-droid. it's still 1.6 there!

PR #439, opened 2015-11-27
fix the bug that modify account name will delete all transaction under the account

fefe982:
quick fix to #436 .

It is caused by REPLACE sql command. REPLACE will delete then insert, and the deletion would cause the transactions to be deleted due to ON DELETE CASCADE on the foreign key constraint.

### DIFF
--- a/app/src/main/java/org/gnucash/android/db/DatabaseAdapter.java
+++ b/app/src/main/java/org/gnucash/android/db/DatabaseAdapter.java
@@ -56,6 +56,10 @@ public abstract class DatabaseAdapter<Model extends BaseModel> {
 
     protected SQLiteStatement mReplaceStatement;
 
+    public enum UpdateMethod {
+        insert, update, replace
+    };
+
     /**
      * Opens the database adapter with an existing database
      * @param db SQLiteDatabase object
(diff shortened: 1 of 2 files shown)


codinguser:
Thanks for the patch.
I was wondering though, since we use replace everywhere, is it more harm than good to have the "delete cascade" constraint in the database? Of course, it is good for ensuring data integrity, but if these side-effects become common....?

fefe982:
I didn't expect this when I first chose the REPLACE statement. I'm working on something to change the replace approach to insert or update / update or insert.

Then one will be able to choose which approach to use when changing the database in the bulk way. Most likely the replace one would not be used after this.

But don't expect it to come out soon.

codinguser:
We can just try to be careful until you finish the approach. So no pressure!
I already modified the database adapters in such a way that the models which contain other models always re-save everything during replace. So that way, we make sure not to lose data in the database. 

One of the benefits of REPLACE over the insert/update approach which we used before is the speed. Insert/Update required at least 2 runs to the database for each operation (much worse for bulk operations). And if we're going to be handling large books going forward (and multiple syncs with users sharing books), then the speed is very important. 

So if any other approach would make it slower, I would say we should stick with REPLACE and deal with the bugs.

fefe982:
I find that now, in a lot of times we do know whether the records exist in the db or not.

E.g, during import, no records exists in the db (which is not the case when REPLACE is introduced), so INSERT is OK. And for this pr, when updating account name, all records should already exist in the db. So UPDATE should be OK.

And something else related to the speed, I found that the db is repeatedly queried when UI changes. I think the query maybe to often, and even thought it should be a bug in the UI. I tried something to resolve this sometime before, but was not successful.
The UI gets a lot of refresh messages. And it seems that it is refreshed twice on creation. Every refresh gets a lot queries. It will refresh when I goes up and down, or swipe left to right. It takes quite some time for the UI to be properly shown.
The balances are calculated repeatedly. They should be able to be cached somewhere.

codinguser:
Great idea. I hadn't thought of it that way. But you're right, we do mostly know when it will an update or insert. As we also know when we're editing/creating transactions and accounts. 

You could change the base DatabaseAdapter interface to use `insert()` and `update()`, `bulkInsert()` and `bulkUpdate()` instead of just an `addRecord()` and `bulkAdd()`. Then just calling the appropriate insert/update methods where applicable.
Looking forward to those changes whenever they're ready  :+1: 

The UI does indeed tend to be refreshed a lot. I will see if I have some redundant refresh calls somewhere and can eliminate them. I'll also look into caching the balances somehow (thanks for the tip)

PR #442, opened 2015-11-29
Keep zh-rCN translated.

lxbzmy:
Keep zh-rCN translated.
I've already upload to https://crowdin.com/project/gnucash-android/zh-CN too.


codinguser:
Great, thanks. I will pull them from Crowd-In.

PR #445, opened 2015-12-03
Transaction amount input uses account commodity

cjwirth:
There was a bug where the amount input view was not having the commodity set. This led to bugs where, if the user creates or edits a transaction in a USD account, but the default is set to JPY, it will
truncate input after the decimal point.

I emailed you about this bug, but I decided that since it's available on Github, I would see if I could fix it myself :smiley_cat:

### DIFF
--- a/app/src/main/java/org/gnucash/android/ui/transaction/TransactionFormFragment.java
+++ b/app/src/main/java/org/gnucash/android/ui/transaction/TransactionFormFragment.java
@@ -494,6 +494,9 @@ public class TransactionFormFragment extends Fragment implements
 		Currency accountCurrency = Currency.getInstance(currencyCode);
 		mCurrencyTextView.setText(accountCurrency.getSymbol());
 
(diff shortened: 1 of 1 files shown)


codinguser:
Thanks. I'll fetch it and merge. 
We have a Trello board full of issues to be fixed. I could add you to it if you'd like to take a swing at some of them ;)

P.S. Send patches to hotfix/patches in the future.

cjwirth:
> We have a Trello board full of issues to be fixed. I could add you to it if you'd like to take a swing at some of them ;)

Sure, why not. I'm an iOS dev by day, but I've done a small bit of Android in the past. Plus I use this app like everyday, so I can take a look at some.

> P.S. Send patches to hotfix/patches in the future.

Sure no problem. Might put that in the contributing section of the readme?

codinguser:
Great. I've added you to the board. 
We do indeed have to improve our contribution information. 

I've merged this patch into hotfix/patches. It will go out with the next patch release.

PR #447, opened 2015-12-05
Improve passcode UI

aksalj:
Tried to improve the look of the passcode lockscreen...

Old | New
----- | ------
old | new


codinguser:
Cool, thanks :+1: 
Would it be possible to reduce the vertical spacing between the buttons so that the overall height of the keypad comes down and the views above don't feel compressed?

codinguser:
On second thought, maybe we should leave it as it is. It has grown on me :D

aksalj:
:+1:

PR #457, opened 2015-12-28
rewrite db update interface

fefe982:
Rewrite addRecord, bulkAddRecord Interface. Now it can specified whether to use INSERT, UPDATE, or REPLACE to update the db.

I've update to calling site of the above methods, and ran only a few tests, but more tests would be needed to make sure it works fine.

Note: in setBindings, the order should be the same as column order in the dbAdapter constructor, and UID always comes last.
Note2: When not sure whether INSERT or UPDATE is appropriate, maybe REPLACE should be used.

P.S. In AccountActivity.init(), PreferenceManager.setDefaultValues(this, R.xml.fragment_transaction_preferences, false); fails. The App can work with this line commented out.
P.S.2 In GnuCashApplication.getDefaultCurrencyCode(),         SharedPreferences prefs = PreferenceManager.getDefaultSharedPreferences(context); fails when the db is empty, e.g. it fails for a newly installed App.


rivaldi8:
I find it a bit cumbersome having to specify the update method for every operation. Also, I'm not sure I'd know which one to specify.

What about making the more appropriate method for each operation the default and then allow to specify a different one if necessary? For example:
```
// by default use DatabaseAdapter.UpdateMethod.insert
mAccountsDbAdapter.addRecord(account);

// if, for instance, DatabaseAdapter.UpdateMethod.update is needed, call
// an overloaded method:
mAccountsDbAdapter.addRecord(account, DatabaseAdapter.UpdateMethod.insert);

// or specify it before:
mAccountsDbAdapter.setUpdateMethod(DatabaseAdapter.UpdateMethod.insert);
mAccountsDbAdapter.addRecord(account); // uses insert
```

Maybe I see it this way because I still don't know much about the database part of  the code, but wanted to give my 2 cents anyway :)

codinguser:
@fefe982 thanks for the pull request. I will test and merge shortly now that holidays are over.

@rivaldi8 inserts are for when you are creating a new record, updates for changing an existing one. Depending on what you are doing it is usually very clear what you want to accomplish. 

I personally would also prefer simple addRecord() and updateRecord() methods which inserted or updated respectively, rather than having to specify an update method as extra parameter. But we can always add these to the base adapter class later and refactor.

fefe982:
If you don't know which one to use, use replace.

PR #475, opened 2016-02-06
Fixes #467 and fixes #464.

### LINKED ISSUE #464: Export Transactions: wrong time suggestion in first export
**How to reproduce**
1. Open a new xml file from GnuCash for Android (GCA);
2. Enter some transaction;
3. Select Export from GCA and the suggested time may be wrong (the difference seems to be exactly your time zone):

**Additional information**
- If we select export again, the suggested time seems to be correct;
- I've installed GCA 2.0.5 on my Nexus 5 running Android 6.0.1.

### LINKED ISSUE #467: Export Transactions: previously imported transactions are exported forever
**How to reproduce**
1. Import a XML file with transactions (I attached a zip file with sample files);
2. Enter some additional transaction;
3. Select Export (pay attention to suggested time, you may have the issue #464);
4. The generated QIF file will have all transactions even the one previously imported.

If we enter another transaction in GCA and export it again the generated QIF file will have the same previously imported transactions plus the new one.

P.S: We cannot import the QIF file in GnuCash desktop as stated at #466

**Additional information**
- I've installed GCA 2.0.5 on my Nexus 5 running Android 6.0.1

**Sample files**
[sample-files.zip](https://github.com/codinguser/gnucash-android/files/86239/sample-files.zip)

alceurneto:
Fixes #467 and fixes #464.


app/src/main/java/org/gnucash/android/util/TimestampHelper.java
```diff
@@ -0,0 +1,99 @@
...
+            DateTimeFormat.forPattern("yyyy-MM-dd HH:mm:ss.SSS");
+
+    /**
+     * Get a {@link String} representing the {@link Timestamp}
+     * in UTC time zone and 'yyyy-MM-dd HH:mm:ss.SSS' format.
+     *
+     * @param timestamp The {@link Timestamp} to format.
+     * @return The formatted {@link String}.
+     */
+    public static String getUtcStringForTimestamp(Timestamp timestamp) {
```
rivaldi8: This name look confusing to me. What about `getUtcStringFromTimestamp()` or `timestampToUtcString()`.

app/src/main/java/org/gnucash/android/util/TimestampHelper.java
```diff
@@ -0,0 +1,99 @@
...
+    /**
+     * Get the {@link Timestamp} with the value of given UTC {@link String}.
+     * The {@link String} should be a representation in UTC time zone with the following format
+     * 'yyyy-MM-dd HH:mm:ss.SSS' OR 'yyyy-MM-dd HH:mm:ss' otherwise an IllegalArgumentException
+     * will be throw.
+     *
+     * @param utcString A {@link String} in UTC.
+     * @return A {@link Timestamp} for given utcString.
+     */
+    public static Timestamp getTimestampForUtcString(String utcString) {
```
rivaldi8: This name look confusing to me. What about `getTimestampFromUtcString()` or `utcStringToTimestamp()`.

app/src/main/java/org/gnucash/android/util/TimestampHelper.java
alceurneto: OK. I'll change `*For*()` to `*From*()` inside `TimestampHelper`.

app/src/main/java/org/gnucash/android/util/TimestampHelper.java
alceurneto: OK.

app/src/main/java/org/gnucash/android/util/PreferencesHelper.java
```diff
@@ -0,0 +1,71 @@
...
+     */
+    private PreferencesHelper() {}
+
+    /**
+     * Tag for logging
+     */
+    protected static final String LOG_TAG = "PreferencesHelper";
+
+
+    private static final String PREFERENCE_LAST_EXPORT_TIME_DEFAULT_VALUE = "PREFERENCE_LAST_EXPORT_TIME_DEFAULT_VALUE";
```
codinguser: @alceurneto Should `PREFERENCE_LAST_EXPORT_TIME_DEFAULT_VALUE` not be a time value rather than a string constant with the same name as the variable? I think you use it later as the default value for a time string right?

alceurneto: Yes. It's used to deal with nothing stored yet. Actually we don't need this
constant. I'll do a little change to make the code more simple and
meaningful.

PR #484, opened 2016-03-22
Bugfix 479: Precision may be lost when saving the exchange rate in the database

### LINKED ISSUE #479: Precision may be lost when saving the exchange rate in the database
When the user enters an exchange rate, it's not saved directly. Instead, the
relation between the original amount and the converted amount is stored. As a
consequence, the exchange rate precision will depend on the precision of the
two currencies.

In addition, the converted amount is (wrongly) stored with the currency of the 
original amount and rounded to its decimal places. If the original amount has
no decimal places, even more precision is lost.

For example, if the user transfers from a Yens account (no decimal places) to an
Euros account (2 decimal places) and enters:
* transaction amount = 10 (Yens, no decimal places)
* exchange rate = 0.123456

GnuCash calculates:

* converted transaction amount: 1 (1.23456 rounded to 0 decimals)
* exchange rate (stored in DB): 1/10 = 0.1

GnuCash should store directly the exchange rate instead of calculating it.

rivaldi8:
Fixes #479


codinguser:
Thanks for the pull request. I'll be reviewing it soon. Just very busy at the moment. 
Cheers,

rivaldi8:
Don't worry, no rush :)

app/src/main/java/org/gnucash/android/model/Money.java
```diff
@@ -199,7 +199,7 @@ private void setCommodity(@NonNull Commodity commodity){
 	 * Sets the commodity for the Money
 	 * @param currencyCode ISO 4217 currency code
 	 */
-	private void setCommodity(@NonNull String currencyCode){
+	public void setCommodity(@NonNull String currencyCode){
```
codinguser: Was just going through the pr and this caught my eye. 
`Money` objects are meant to be immutable, just like Java Strings. Could you change this back to private and use `Money.withCurrency()` at the relevant places instead?

rivaldi8:
Fixed the issue with the immutability of Money.

codinguser:
Gracias

app/src/main/java/org/gnucash/android/model/Price.java
```diff
@@ -78,6 +95,7 @@ public void setType(String type) {
     }
 
     public long getValueNum() {
+        reduce();
```
codinguser: This is causing the valueNum of the price to be halved when it is retrieved. 
What is the purpose of this `reduce()` call here?

PR #489, opened 2016-04-10
Avoid returning nulls from some Account methods

rivaldi8:
I've applied the idea of the [Null object pattern](https://en.wikipedia.org/wiki/Null_Object_pattern) to some methods of Account. It simplifies the code of its clients by not having to check for null values.

If it's ok, I plan to send more small pull requests like this one, trying to simplify the code little by little. This would make the code easier to understand, as sometimes I have difficulties due to the many conditionals.


codinguser:
Sure, send them! :+1:

PR #490, opened 2016-04-11
Fix saving state in AccountsListFragment

chiken88:
AccountsListFragment didn't save it state that caused to displaying identical fragments in ViewPager.

Steps to reproduce:
- open homescreen (AccountsActivity);
- check ViewPager fragments;
- rotate device;
- check ViewPager fragments, they all will have mDisplayMode with value TOP_LEVEL.

device-2016-04-11-231446


codinguser:
👍

Issue #491, opened 2016-04-22
Make GnuCash Android available outside of Google store


strk:
Could you please make the .apk available on the official website for direct download ?
Using GooglePlay requires a Google account which not everyone is willing to register.
Thanks in advance.

codinguser:
https://github.com/codinguser/gnucash-android/releases

strk:
Thanks! See #493 for giving this information in the README (which I did read :)

PR #493, opened 2016-04-22
Add GitHub releases as a way to obtain the application

strk:
See #491


codinguser:
Sorry took a while to process this. 
Pull requests should be issued against the develop branch. I guess you figured it out. 
We need to add a contribution guide to the wiki.

Issue #196, opened 2014-09-06
Need ability to delete transactions


twunder:
GnuCash for Android does not appear to be able to delete a single transaction. Previously, the only way to delete a transaction was to edit it and make the amount 0. Since transaction editing is broken in 1.4.2 (see previous incident), that is not an option.

Even after editing of transactions is fixed, it would be much better if a transaction could be deleted from the UI instead of being zeroed out.

PR #498, opened 2016-04-24
Avoid Price.toString() failing for certain values of numerator/denominator.

rivaldi8:
Fixes a regression introduced in 66cc533 (bugfix #479) in which the numerator/denominator division fails with an `ArithmeticException` if no `MathContext` is provided. Somehow I missed it when moving code around.

Fixes Crashlitics issue [#196](https://fabric.io/gnucash/android/apps/org.gnucash.android/issues/571b4a95ffcdc04250f98ba03)

### DIFF
--- a/app/src/main/java/org/gnucash/android/model/Price.java
+++ b/app/src/main/java/org/gnucash/android/model/Price.java
@@ -4,6 +4,7 @@ import org.gnucash.android.util.TimestampHelper;
 
 
 import java.math.BigDecimal;
+import java.math.MathContext;
 import java.sql.Timestamp;
 import java.text.DecimalFormat;
 import java.text.NumberFormat;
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -4,7 +4,7 @@ buildscript {
 		jcenter()
 	}
 	dependencies {
-		classpath 'com.android.tools.build:gradle:1.5.0'
+		classpath 'com.android.tools.build:gradle:2.0.0'
```
codinguser: @rivaldi8  Thanks for the pr. 
This version upgrade breaks the build, because it depends on a higher version of gradle. 
We will upgrade for the 2.1.x series. But 2.0.x should stay with build tools 1.5.0

rivaldi8:
Ok, I've updated the pull request dropping the commit.

PR #509, opened 2016-06-20
Replace uses of Currency with Commodity in Account

rivaldi8:
This pull request is a step towards removing use of Currency, at least for the cases where Commodity is more appropriate. More PRs will be needed to complete the removal from other classes. This will avoid duplication of information and confusion caused by the use of two similar classes.

### DIFF
--- a/app/src/main/java/org/gnucash/android/ui/report/ReportsOverviewFragment.java
+++ b/app/src/main/java/org/gnucash/android/ui/report/ReportsOverviewFragment.java
@@ -35,7 +35,6 @@ import com.github.mikephil.charting.data.PieData;
 import com.github.mikephil.charting.data.PieDataSet;
 
 import org.gnucash.android.R;
-import org.gnucash.android.app.GnuCashApplication;
 import org.gnucash.android.db.adapter.AccountsDbAdapter;
 import org.gnucash.android.model.Account;
 import org.gnucash.android.model.AccountType;
@@ -49,9 +48,7 @@ import org.joda.time.LocalDate;
 
 import java.util.ArrayList;
 import java.util.Collections;
-import java.util.Currency;
 import java.util.List;
-import java.util.Locale;
 
 import butterknife.Bind;
 import butterknife.OnClick;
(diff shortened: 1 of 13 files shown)


app/src/main/java/org/gnucash/android/ui/report/ReportsOverviewFragment.java
```diff
@@ -178,14 +175,13 @@ protected void generateReport() {
...
     private PieData getData() {
-        String mCurrencyCode = GnuCashApplication.getDefaultCurrencyCode();
         PieDataSet dataSet = new PieDataSet(null, "");
         List<String> labels = new ArrayList<>();
         List<Integer> colors = new ArrayList<>();
         for (Account account : mAccountsDbAdapter.getSimpleAccountList()) {
             if (account.getAccountType() == AccountType.EXPENSE
                     && !account.isPlaceholderAccount()
-                    && account.getCurrency() == Currency.getInstance(mCurrencyCode)) {
+                    && account.getCommodity() == mCommodity) {
```
codinguser: Java Currency class uses Singletons, so the `==` operator will work for them. But I don't think Commodity works this way at the moment. It does not use singletons
That means the `equals()` method should probably be used here instead. It has been properly implemented for the Commodity class.

app/src/main/java/org/gnucash/android/ui/report/barchart/StackedBarChartFragment.java
```diff
@@ -163,7 +162,7 @@ protected BarData getData() {
             for (Account account : mAccountsDbAdapter.getSimpleAccountList()) {
                 if (account.getAccountType() == mAccountType
                         && !account.isPlaceholderAccount()
-                        && account.getCurrency() == mCurrency) {
+                        && account.getCommodity() == mCommodity) {
```
codinguser: Same here, may not have the effect you expect. use `equals()` instead

app/src/main/java/org/gnucash/android/ui/report/piechart/PieChartFragment.java
```diff
@@ -152,7 +152,7 @@ private PieData getData() {
         for (Account account : mAccountsDbAdapter.getSimpleAccountList()) {
             if (account.getAccountType() == mAccountType
                     && !account.isPlaceholderAccount()
-                    && account.getCurrency() == mCurrency) {
+                    && account.getCommodity() == mCommodity) {
```
codinguser: Use `equals()`

codinguser:
Thanks for the pr @rivaldi8 . I made some comments inline. Please fix and I'll merge. 

Also, I will probably look to including an IntelliJ style file so that we can all always use the same spaces/tabs for this project. Would make it easier to review code without all the whitespace changes.

rivaldi8:
You are never careful enough :) Well, I've updated the PR to use equals().

The InellilJ style file would be nice. Sometimes it's annoying having to revert tab replaces done automatically by the IDE.

PR #511, opened 2016-06-24
Enable remote debugging from Chrome developer tools

rivaldi8:
Adds [Stetho](http://facebook.github.io/), a debug bridge that, among other things, allows to access the database and preferences of the application through Chrome developer tools.

The database access is especially useful. It allows to browse the tables and send SQL queries.

To make it simple, I've included the library in all builds, although at runtime it's only enabled for debug builds. If you find it useful and want it to be included only for debug builds, I can set it up with a separate debug Application like explained [here](http://stackoverflow.com/a/30172663).

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -228,6 +228,8 @@ dependencies {
         transitive = true;
     }
 
+    compile 'com.facebook.stetho:stetho:1.3.1'
+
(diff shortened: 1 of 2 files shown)


codinguser:
This sounds like a good tool to have for developing. 
In the future, we should probably look into having it only included in the debug builds.

PR #524, opened 2016-07-21
Bugfix 506: Total balance is incorrect for one of my accounts

### LINKED ISSUE #506: Total balance is incorrect for one of my accounts
For one of my accounts the total balance is not correct. At the moment the total balance should be 1656.01BGN but the total balance shown is 1714.77BGN. There are no future transactions. I actually went through all transactions and calculated them by hand and came up with the exact same number as is shown in my actual bank account: 1656.01BGN.

I'm managing 9 accounts but this seems to be the only one exhibiting the problem. What is special about this account is that it is in BGN, but it contains several transactions that are done between this account and a couple of my other accounts which are in EUR. Possibly something has gone wrong in the currency conversion.

I'm willing to privately share a backup of my data files if this could help in debugging the problem.

rivaldi8:
Fixes #506. Does also some refactorings.

### DIFF
--- a/app/src/main/java/org/gnucash/android/ui/transaction/TransactionFormFragment.java
+++ b/app/src/main/java/org/gnucash/android/ui/transaction/TransactionFormFragment.java
@@ -24,6 +24,7 @@ import android.content.res.Configuration;
 import android.database.Cursor;
 import android.inputmethodservice.KeyboardView;
 import android.os.Bundle;
+import android.support.annotation.NonNull;
 import android.support.v4.app.Fragment;
 import android.support.v4.widget.SimpleCursorAdapter;
 import android.support.v7.app.ActionBar;
@@ -635,13 +636,23 @@ public class TransactionFormFragment extends Fragment implements
         intent.putExtra(UxArgument.SELECTED_ACCOUNT_UID, mAccountUID);
         intent.putExtra(UxArgument.AMOUNT_STRING, baseAmountString);
         if (mSplitsList != null) {
-            ArrayList<String> splitStrings = new ArrayList<>();
-            for (Split split : mSplitsList) {
-                splitStrings.add(split.toCsv());
-            }
-            intent.putStringArrayListExtra(UxArgument.SPLIT_LIST, splitStrings);
+            intent.putStringArrayListExtra(UxArgument.SPLIT_LIST, getSplitsAsCsvList());
         }
         startActivityForResult(intent, REQUEST_SPLIT_EDITOR);
+    }
+
+
+    /**
+     * Returns the list of splits as a new list with the splits converted to CSV.
+     *
+     * @return the list of splits as a new list with the splits converted to CSV.
+     */
+    private @NonNull ArrayList<String> getSplitsAsCsvList() {
+        ArrayList<String> splitStrings = new ArrayList<>();
+        for (Split split : mSplitsList) {
+            splitStrings.add(split.toCsv());
+        }
+        return splitStrings;
     }
 
     /**
--- a/app/src/main/java/org/gnucash/android/ui/transaction/TransactionFormFragment.java
+++ b/app/src/main/java/org/gnucash/android/ui/transaction/TransactionFormFragment.java
@@ -56,14 +56,12 @@ import com.codetroopers.betterpickers.recurrencepicker.EventRecurrenceFormatter;
 import com.codetroopers.betterpickers.recurrencepicker.RecurrencePickerDialogFragment;
 
 import org.gnucash.android.R;
-import org.gnucash.android.app.GnuCashApplication;
-import org.gnucash.android.db.adapter.CommoditiesDbAdapter;
 import org.gnucash.android.db.DatabaseSchema;
 import org.gnucash.android.db.adapter.AccountsDbAdapter;
 import org.gnucash.android.db.adapter.DatabaseAdapter;
 import org.gnucash.android.db.adapter.ScheduledActionDbAdapter;
 import org.gnucash.android.db.adapter.TransactionsDbAdapter;
-import org.gnucash.android.model.AccountType;
+import org.gnucash.android.model.Account;
 import org.gnucash.android.model.Commodity;
 import org.gnucash.android.model.Money;
 import org.gnucash.android.model.Recurrence;
@@ -218,17 +216,11 @@ public class TransactionFormFragment extends Fragment implements
      */
     private Calendar mTime;
 
-    /**
-     * The AccountType of the account to which this transaction belongs.
-     * Used for determining the accounting rules for credits and debits
-     */
-    AccountType mAccountType;
-
 
     private String mRecurrenceRule;
     private EventRecurrence mEventRecurrence = new EventRecurrence();
 
-    private String mAccountUID;
+    private Account mAccount;
 
     private List<Split> mSplitsList = new ArrayList<>();
 
@@ -261,7 +253,7 @@ public class TransactionFormFragment extends Fragment implements
      * Starts the transfer of funds from one currency to another
      */
     private void startTransferFunds() {
-        Currency fromCurrency = Currency.getInstance(mTransactionsDbAdapter.getAccountCurrencyCode(mAccountUID));
+        Currency fromCurrency = Currency.getInstance(mAccount.getCommodity().getCurrencyCode());
         long id = mTransferAccountSpinner.getSelectedItemId();
         String targetCurrency = mAccountsDbAdapter.getCurrencyCode(mAccountsDbAdapter.getUID(id));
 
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/gnucash/android/ui/transaction/TransactionFormFragment.java
```diff
@@ -745,15 +741,15 @@ private void saveNewTransaction() {
 
         mAccountsDbAdapter.beginTransaction();
```
rivaldi8: This is confusing to me. Why the  mAccountsDbAdapter.beginTransaction() if we are adding records to mTransactionsDbAdapter?

rivaldi8:
I've updated the pull request to fix some issues that slip through.

app/src/main/java/org/gnucash/android/ui/transaction/TransactionFormFragment.java
codinguser: `#beginTransaction()` works on the whole database, not on specific tables. So it makes no difference.

codinguser:
@rivaldi8  Thanks for the PR. 
Could you please add some tests to it that demonstrate that the problem is fixed? 

Also it would be good to send bugfixes and refactorings as separate pull requests, it makes it easier to review changes, giving more attention to bugfixes than simple refactorings (which mostly can be merged immediately).

rivaldi8:
Ok! However, i won't be able to do so until next week, when I'm back from
holidays. That's why i sent the pr a bit in a hurry.

codinguser:
No problem. Enjoy your holiday :)

codinguser:
I was trying to write a test for this pull request and I realized that there is another issue in the app. It can be reproduced as follows: 
- Create two accounts A1 in USD and A2 in EUR
- Create a multi-currency transaction in A1 with an amount of $10. 
- Enter converted amount €20 in the Transfer dialog and save the transaction
- Open the same newly created transaction for editing, but do not change anything
- Save it. 
- Now the transaction will show two splits of $10 and €10 (instead of $10 and €20)

I will try to make a fix for this. I have the feeling it might fix the #506  issue as well.

Also, I cannot merge this because this pull request introduces another bug. When I try to create a simple multi-currency transaction, I end up with 4 splits (instead of 2) - one duplicate and another to an imbalance account. I'm not sure of what other side effects there may be as I've not tested further.

Since you're on holiday, I will try to just work on the fixes directly on top of `develop`.

rivaldi8:
I guess this other issue you found is from some rare code path. This class is complex and has a lot of code paths that may hide other subtle bugs like these. That's why I tried to simplify with the refactoring.

I had just started this afternoon to write the test too, but if you are on it, I'd rather leave it to you. Since I got back on Sunday I've been trying to get the instrumentation tests to run with no luck. On the on hand, for some reason my phone has always failed to run the tests and haven't been able to fix it now either. On the other, the Android emulator refuses to run since some time ago. So writing instrumentation tests without being able to run them is difficult to get right.

I've also found some bug introduced by this pull request, I guess from the refactorings. So just pick the first commit and discard the rest.

codinguser:
You are right, that code logic for handling the saving of transactions is
really a big tangle with multiple code paths. But that is due to all the
different possibilities for creating transactions, so it is hard to
simplify it. I hope to eventually write tests to cover most of the cases
and then we can maybe try to refactor it somehow just to make it clearer.

What error do you get when you run the instrumentation tests? It would be
good if we could get them working for you.

Which one do you mean as the first
commit? 2d76e1f00ddd684179f4b97b54340eccb63a6daa
or fc99a5a71b3e0b954cbca0a57da6c2f24a2332e8
Also, please separate pull requests refactorings and bug fixes in the
future. That way it is easier and quicker to merge the refactorings while
paying more attention to the bugfixes.

rivaldi8:
I think part of the complexity comes from the fact that Transaction is more a data holder than full class with responsibilities. From the UI we deal with all the logic of splits. We create and set them up, and then we add them to the transaction, which is tricky and error prone. In general we have too much business logic entangled in UI classes, which makes it difficult to test.

In the case of Transaction maybe we could have something like `Transaction.addSplit(memo, amount)` which would automatically set its account and type, and could also add the pair split. Or maybe use the Builder pattern or add a business layer. I'm not sure, but somehow we should move as much of this code as possible out of the UI, so it can be more easily tested from plain unit tests.

The error I get when trying to run the tests is this:
```
nstalling APK 'GnucashAndroid_v2.1.0-dev0_rfc99a5a.apk' on 'Aquaris E4 - 5.0' for app:developmentDebug
12:17:08 E/918313569: Error while uploading GnucashAndroid_v2.1.0-dev0_rfc99a5a.apk : Unknown failure ([CDS]close[0])
Unable to install /home/alex/program/gnucash-android/gnucash-android/app/build/outputs/apk/GnucashAndroid_v2.1.0-dev0_rfc99a5a.apk
com.android.ddmlib.InstallException: Failed to install all 
        at com.android.ddmlib.SplitApkInstaller.install(SplitApkInstaller.java:91)
        at com.android.ddmlib.Device.installPackages(Device.java:901)
        at com.android.builder.testing.ConnectedDevice.installPackages(ConnectedDevice.java:119)
        at com.android.build.gradle.internal.tasks.InstallVariantTask.install(InstallVariantTask.java:134)
[...]
```
It fails to install the APK. If I install it manually and try to run the tests, it fails because it doesn't find the tests:
```
adb -d shell am instrument -w org.gnucash.android.devel/org.gnucash.android.test.ui.util.GnucashAndroidTestRunner
INSTRUMENTATION_STATUS: id=ActivityManagerService
INSTRUMENTATION_STATUS: Error=Unable to find instrumentation info for: ComponentInfo{org.gnucash.android.devel/org.gnucash.android.test.ui.util.GnucashAndroidTestRunner}
INSTRUMENTATION_STATUS_CODE: -1
android.util.AndroidException: INSTRUMENTATION_FAILED: org.gnucash.android.devel/org.gnucash.android.test.ui.util.GnucashAndroidTestRunner
```
It looks like the classes aren't included. Or maybe I'm doing something wrong. I'd say there is something wrong with this brand/model. My partner has the same model and gives the same error. Other phones I've tested run the tests correctly. I'll keep investigating...

I meant the commit 2d76e1f, the one that fixes the bug.

Yes, I'll send separate pull requests from now on.

codinguser:
I agree that there is a lot of control logic in the UI at the moment. We can begin to explore ways to extract this slowly (and after adding tons of tests ;) ). Maybe something like a Model-View-Controller, with the Controllers being missing at the moment.  

At the beginning I tried to do a hard separation, model classes being simply data holders, and then Android-specific (like database and UI) components separate. But it still got mixed over time. 

The main issue is that the UI is live and dynamic, we respond to user inputs, pre-process, and validate stuff before saving. Especially for transactions and splits. We can look into how to separate this validation into Controllers. But they would still need deep access to the UI. I'm very open to ideas on how to achieve this.

From the output you provided, It really does seem as though the tests not working are a device-specific issue. But it makes me wonder why they don't work in your emulator either.... (I'm assuming you have HyperV disabled if using Windows). 

On the other hand, Robolectric can also be used to test the UI (see example here: http://robolectric.org/). We already use Robolectric, so it is just a matter of writing the tests. I would actually prefer many more Robolectric tests since they run fast and are easily supported on Travis CI. So you could definitely write your tests that way. Then we will have just a few tests for checking main workflows that will require a device or emulator and are written in Espresso.

rivaldi8:
Yes, I'm already experimenting with ways to accomplish this separation. The refactorings here are part of it, but of course it's complicated and very easy to add bugs, even after careful testing. I'll keep experimenting and propose something when I have a more clear idea about it.

I'm using Linux, so it's not HyperV. I think it broke after the last Ubuntu upgrade, but as I barely used it (to slow) I'm not sure.

Oh! I didn't know about Robolectric. I'll have a look at it.

codinguser:
I was thinking a bit again about the separation of UI and backend logic and I researched a bit. It seems on Android the paradigm is that Activities and Fragment classes should hold the glue between models and the views. They represent the Controllers in the MVC pattern and the inflatable widgets are the views. It is therefore unavoidable that they have lots of code that touches front and back. 
I think what we can do is work to make that code better, simpler, easier to manage. the 'saveNewTransaction()' method was too convoluted and I have tried to simplify it a bit, while keeping things working. 

But if you think up any interesting ways to separate UI from logic, I would love to see it :)

rivaldi8:
Yes, I seem to recall having read something similar about how Android is supposed to be used in this regard.

I don't have any clear idea of how to separate UI from logic yet. I'm reading some Android books to learn about all components of which I just have a vague idea of what they are. So right now, my head is just chaos. I'll need some time to put all in order :)

For now, I think just simplifying and pushing code to *DbAdapters and model classes should be enough.

PR #527, opened 2016-08-04
Bugfix 526: Opening a scheduled transaction to view causes app to crash

### LINKED ISSUE #526: Opening a scheduled transaction to view causes app to crash
The scheduled transaction list is displayed, but the individual items cannot be opened for editing.. 
This behaviour currently occurs on `develop` branch.

This can be fixed and extended so that tapping on a scheduled transaction does not directly go to editing it, but rather first displays details of it using `TransactionDetailFragment`. And then from there, it can be edited further.

rivaldi8:
Fixes #526 

I haven't implemented going through `TransactionDetailFragment` first yet. I've found some bugs in the way that I'd like to fix first.

By the way, I've noticed the start and end periods are stored both in `scheduled_actions` and `recurrences` tables. Is this duplicated for convenience or should we remove one of them?

### DIFF
--- a/app/src/main/java/org/gnucash/android/model/Recurrence.java
+++ b/app/src/main/java/org/gnucash/android/model/Recurrence.java
@@ -263,6 +263,9 @@ public class Recurrence extends BaseModel {
      * @return Number of occurrences, or -1 if there is no end date
      */
     public int getCount(){
+        if (mPeriodEnd == null)
+            return -1;
+
         int count = 0;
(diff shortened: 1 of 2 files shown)


codinguser:
Thanks @rivaldi8 . 
yeah, it is better to fix the bugs first and then the other "feature" stuff later. 

The duplicate entries in the tables are done on purpose. I cannot remember the exact reason now, but I know I did that on purpose :)

rivaldi8:
Ok, thanks for the clarification! :)

PR #530, opened 2016-08-04
Fix setting Recurrence.mPeriodEnd from ScheduledAction.setEndTime()

rivaldi8:
It looks like a copy-paste mistake. I doesn't seem there was any bug caused by this.


rivaldi8:
I've added a commit with a small refactoring, otherwise I'd have to wait for you to merge this pull request and then send the another pull request depending on this one. I hope it's ok :)

codinguser:
It's fine, thanks :)

PR #531, opened 2016-08-04
Small ScheduledActionTest refactor

rivaldi8:
Depends on #530


codinguser:
What happened here? :)

rivaldi8:
I selected the wrong base branch :) Maybe you could set the default branch to `develop` or `hotfix/patches` (see [Setting the default branch](https://help.github.com/articles/setting-the-default-branch/)). What do you think?

codinguser:
I'll think about it. 
I considered it before, but thought it might cause people who just checkout to build the app for themselves to build a develop build. And those can be in a not so stable state.

rivaldi8:
Yes, you are right. There should be a separate option for pull requests.

PR #549, opened 2016-08-19
Bugfix 528: When using "XXX no currency" the accounts balances aren't calculated

### LINKED ISSUE #528: When using "XXX no currency" the accounts balances aren't calculated
When using "XXX no currency", if some transactions are added to an account, the account and its parents are still listed with 0 balance.

Received through User Voice.

rivaldi8:
This fixes #528, however, it does so by removing an explicit check to not calculate the balance for this currency. There was no explanation in the code nor in the commit message. Also, I can't think of any reason to exclude this currency. Any idea @codinguser?

The check was added in ff7f8c0. @fefe982, if you could shed some light on this, it would be highly appreciated :)

### DIFF
--- a/app/src/main/java/org/gnucash/android/db/adapter/SplitsDbAdapter.java
+++ b/app/src/main/java/org/gnucash/android/db/adapter/SplitsDbAdapter.java
@@ -220,8 +220,7 @@ public class SplitsDbAdapter extends DatabaseAdapter<Split> {
                 long amount_denom = cursor.getLong(1);
                 String commodityCode = cursor.getString(2);
                 //Log.d(getClass().getName(), commodity + " " + amount_num + "/" + amount_denom);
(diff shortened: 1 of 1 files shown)


codinguser:
@rivaldi8 IIRC the reason for that was because currently we assign XXX to any commodity we do not know when importing. This means stocks, tradings and lots of other stuff which GnuCash desktop has which we don't. (Users can also define custom commodities).
 
Since XXX basically means "unknown", we do not assume to know how to calculate a balance from that.
@fefe982 I am right in my recollection?

rivaldi8:
Ok, I'll ask the user that reported the issue about how he uses this currency and see if something can be done.

rivaldi8:
I got no response from the user.

hughdavenport:
I ran into this issue. Just changed to XTS instead which shows balances and unlikely to conflict

PR #563, opened 2016-08-28
Bugfix 562: Scheduled transaction imported with imbalanced splits

### LINKED ISSUE #562: Scheduled transaction imported with imbalanced splits
From user report:

> I've come across a problem with scheduled actions. On GnuCash on my Linux desktop I have several scheduled actions that run every month. On android though so of them are missing the one of the amounts. For example i have £10 going from my bank account to a savings account. On android when I look at the scheduled action it has £0 coming out of my bank and £10 going into the savings account, leaving the transaction -£10 imbalanced.

rivaldi8:
Fixes #562

### DIFF
--- a/app/src/main/java/org/gnucash/android/importer/GncXmlHandler.java
+++ b/app/src/main/java/org/gnucash/android/importer/GncXmlHandler.java
@@ -1050,11 +1050,16 @@ public class GncXmlHandler extends DefaultHandler {
      */
     private void handleEndOfTemplateNumericSlot(String characterString, TransactionType splitType) {
         try {
-            BigDecimal amountBigD = GncXmlHelper.parseSplitAmount(characterString);
-            Money amount = new Money(amountBigD, getCommodityForAccount(mSplit.getAccountUID()));
-            mSplit.setValue(amount.abs());
-            mSplit.setType(splitType);
-            mIgnoreTemplateTransaction = false; //we have successfully parsed an amount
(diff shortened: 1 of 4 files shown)


codinguser:
@rivaldi8  This merge is failing on Travis CI. which is strange because it passed on it's own
Was your develop up-to-date when you opened this pr?

codinguser:
ok, seems like many other things are failing too. I'll investigate

rivaldi8:
@codinguser Yes, ensured the branch was updated before pushing.

I always have a test failing because of another setting the locale. Now that one no longer fails and fails this one instead. I guess that's why I didn't notice.

codinguser:
:) interesting 
I would like to know which (of the unit) tests fail because of locale. I would prefer to make the tests locale-independent.

rivaldi8:
The locale dependent test was AccountsDbAdapterTest.importingXml_shouldSetDefaultCurrency, which you fixed in e5daf4b. So all unit tests succeed now, at least for me :)

codinguser:
Awesome! 👍🏾😃






On Wed, Aug 31, 2016 at 11:28 PM +0200, "Àlex Magaz Graça" <notifications@github.com> wrote:












The locale dependent test was AccountsDbAdapterTest.importingXml_shouldSetDefaultCurrency, which you fixed in e5daf4b. So all unit tests succeed now, at least for me :)

PR #578, opened 2016-09-12
Bugfix 577: Can't save transactions on single entry mode

rivaldi8:
Introduced in 96a2045. I guess you meant this condition instead.

I haven't added a tests as, apart from not being able to run them on my phone yet, I'd say `TransactionsActivityTest.testAutoBalanceTransactions` should fail.


codinguser:
Thanks for catching that. Do you know if this affects the current release?

codinguser:
oh, sorry, just saw that this was a hotfix patch 👍

PR #585, opened 2016-09-16
Bugfix 583:   Scheduled exports are always executed twice a day

### LINKED ISSUE #583: Scheduled exports are always executed twice a day
#### Expected behaviour
Scheduled exports are executed according to the frequency set (daily, weekly...).

#### Actual behaviour
Scheduled exports are always executed twice a day, independently of the actual frequency set.

1. On the application menu, select _Export_.
2. Select _SD card_ as destination.
3. Press _Tap to create schedule_.
4. Select
 * REPEAT WEEKLY
 * Monday
 * Always
5. Press _Accept_
6. Press _Export_

Every day from now on, transactions will be exported instead of once per week on Mondays.

#### Software specifications
* GnuCash Android version: 2.1.1
* System Android version: 5.0
* Device type: bq Aquaris E4

rivaldi8:
Fixes #583.

### DIFF
--- a/app/src/main/java/org/gnucash/android/service/ScheduledActionService.java
+++ b/app/src/main/java/org/gnucash/android/service/ScheduledActionService.java
@@ -164,6 +164,9 @@ public class ScheduledActionService extends IntentService {
         if (endTime > 0 && endTime < now)
             return executionCount;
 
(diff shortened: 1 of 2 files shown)


codinguser:
Gracias

PR #589, opened 2016-09-21
Develop

janmercl:


janmercl:
1. 

#### Expected behaviour

- on mobile device (Android 7.0, Nexus 5X) create Book2, which creates default Accounts (Assets, Equity, Expenses, Income and Liabilities with their sub-accounts)

- create the transactions in mobile device as follows:
Expenses:Auto:Gas 56,00€ (expense) -> Assets:Current Assets:Cash in Wallet

- export the transactions as .qif file type to Google Drive

- download to PC (linux Mint 18)

- import to gnuCash (2.6.12) with no accounts - they are automatically created during the transactions import

- after import expecting in PC program:
Assets: -56,00€
Expenses: 56,00€


#### Actual behaviour

- after import 
in PC program:
Assets: 56,00€
Expenses: 56,00€

- found a wrong Account type: Assets, account type 'Income'

#### Software specifications
* GnuCash Android version: 2.1.1
* System Android version: 7.0
* Device type: Nexus 5X

codinguser:
Hi,
What is your device locale?
The default accounts are imported based on your locale. So if there is an
error in the template for you locale, please let me know so we can fix it.
Cheers,

Issue #591, opened 2016-10-03
Scheduled exports run with each service run until missed executions are recovered


rivaldi8:
#### Steps to reproduce the behaviour
1. Create a scheduled export.
2. Miss some executions (by turning off the phone, for example).

#### Expected behaviour
The scheduled export isn't executed until the next planned execution.

#### Actual behaviour
The scheduled export is executed every time the ScheduledActionService runs until the missed executions have been done. For example, if two executions have been missed, two exports will be done, then it won't be run again until the next planned date.

#### Software specifications
* GnuCash Android version: 2.1.2
* System Android version: 5.0
* Device type: bq Aquaris E4

codinguser:
Is this still an issue in the current code base in `hotfix/patches`? 
I vaguely remember running into this and fixing it. But I may be wrong

rivaldi8:
I think you refer to #583 which is similar but not the same. There's still another bug related with scheduled action that I'll report shortly.

rivaldi8:
I think you refer to #583 which is similar but not the same. There's still another bug related with scheduled action that I'll report shortly.

rivaldi8:
Fixed in hotfix/patches.

PR #593, opened 2016-10-04
Bugfix 591:  Scheduled exports run with each service run until missed executions are recovered

rivaldi8:
@codinguser please, review this as it's hackish and I might have got something wrong. I have another bug depending on this one so I'll be working on it meanwhile.

Fixes https://github.com/codinguser/gnucash-android/issues/591

### DIFF
--- a/app/src/main/java/org/gnucash/android/model/ScheduledAction.java
+++ b/app/src/main/java/org/gnucash/android/model/ScheduledAction.java
@@ -313,7 +313,12 @@ public class ScheduledAction extends BaseModel{
     }
 
     /**
-     * Returns how many times this scheduled action has already been executed
+     * Returns how many times this scheduled action has already been executed.
+     *
+     * <p><b>Note:</b> the number returned here will include not only the
+     * real execution count but the missed ones too. This is because
+     * computeNextScheduledExecutionTime depends on it to give the correct date.</p>
+     *
      * @return Number of times this action has been executed
      */
     public int getExecutionCount(){
@@ -322,6 +327,11 @@ public class ScheduledAction extends BaseModel{
 
     /**
      * Sets the number of times this scheduled action has been executed
+     *
+     * <p><b>Important:</b> when updating this field include not only the real
+     * execution count but the missed executions too. This is because
+     * computeNextScheduledExecutionTime depends on it to give the correct date.</p>
+     *
      * @param executionCount Number of executions
      */
     public void setExecutionCount(int executionCount){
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/gnucash/android/service/ScheduledActionService.java
```diff
@@ -175,14 +175,32 @@ private static int executeBackup(ScheduledAction scheduledAction, SQLiteDatabase
...
             //wait for async task to finish before we proceed (we are holding a wake lock)
             new ExportAsyncTask(GnuCashApplication.getAppContext(), db).execute(params).get();
-            scheduledAction.setExecutionCount(++executionCount);
+            ++executionCount;
         } catch (InterruptedException | ExecutionException e) {
             Crashlytics.logException(e);
             Log.e(LOG_TAG, e.getMessage());
         }
+
+        // HACK: Add missed runs to executionCount so computeNextScheduledExecutionTime()
```
codinguser: If this fixes the problem, then we can live with it for now. As long as the meaning of the executionCount is not changed.
In the future backups & exports may be separated from scheduled transactions and then we will have to revisit this.

app/src/main/java/org/gnucash/android/model/ScheduledAction.java
```diff
@@ -322,6 +327,11 @@ public int getExecutionCount(){
 
     /**
      * Sets the number of times this scheduled action has been executed
+     *
+     * <p><b>Important:</b> when updating this field include not only the real
```
rivaldi8: Yes, I agree. It was a last minute change and I really don't like it either. Specially taking into account that it only applies to scheduled exports.

app/src/test/java/org/gnucash/android/test/unit/service/ScheduledActionServiceTest.java
```diff
@@ -280,10 +280,25 @@ public void recurringTransactions_shouldHaveScheduledActionUID(){
...
+     * useful, so just one should be done.</p>
+     *
+     * <p><i>Note</i>: the execution count will include the missed runs
+     * as computeNextScheduledExecutionTime depends on it.</p>
+     */
     @Test
     public void scheduledBackups_shouldRunOnlyOnce(){
         ScheduledAction scheduledBackup = new ScheduledAction(ScheduledAction.ActionType.BACKUP);
-        scheduledBackup.setStartTime(new DateTime(2016, 2, 17, 17, 0).getMillis());
+        scheduledBackup.setStartTime(LocalDateTime.now()
```
rivaldi8: Yes, it really nice to test for certain time-dependant conditions :)

app/src/main/java/org/gnucash/android/service/ScheduledActionService.java
rivaldi8: Yes, my initial intention was to separate scheduled exports and transactions into different classes but again, it depended on having an XML pull parser for importing. That's why I did this temporal hack. So I guess I'll start with the parser as soon as I fix these bugs.

codinguser: We'll separate them with bigger db schema changes down the line.

Issue #184, opened 2014-08-23
On the widget, negatives values are green in 1.4.0


johngray1965:
On the widget, negative values are now green, they used to be red as I would have expected.

PR #597, opened 2016-10-09
Fix widgets with negative values being green

czan:
The number used to calculate whether a widget is red or green should be the same value as the one that is displayed. It was instead calling a (buggy) method which calculated a different balance to the one displayed. This patch fixes it to determine the colour using the same value as the one it is displaying.

This should fix issue #184.

### DIFF
--- a/app/src/main/java/org/gnucash/android/ui/homescreen/WidgetConfigurationActivity.java
+++ b/app/src/main/java/org/gnucash/android/ui/homescreen/WidgetConfigurationActivity.java
@@ -175,7 +175,7 @@ public class WidgetConfigurationActivity extends Activity {
 
 		views.setTextViewText(R.id.transactions_summary,
 				accountBalance.formattedString(Locale.getDefault()));
-		int color = account.getBalance().isNegative() ? R.color.debit_red : R.color.credit_green;
+		int color = accountBalance.isNegative() ? R.color.debit_red : R.color.credit_green;
 		views.setTextColor(R.id.transactions_summary, context.getResources().getColor(color));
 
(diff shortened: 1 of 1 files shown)


codinguser:
Thanks for the patch. The build fails for some reason with this patch, even though the changes seem unrelated to the failing tests. 
I'd like to figure out why before I merge it.

czan:
When I run the tests in the `develop` branch they fail in the same way. Should I be pulling into a different branch?

czan:
I found the error in the tests!

In `ScheduledActionTest#testComputingNextScheduledExecution` and `ScheduledActionTest#testComputingTimeOfLastSchedule` you have a line `periodType.setMultiplier(2);`. In one case it's affection `PeriodType.MONTH`, and the other `PeriodType.WEEK`. The problem is that both of those are **global** enum values. That means that when the failing tests (`RecurrenceTest#settingCount_shouldComputeCorrectEndTime` and `RecurrenceTest#testRecurrenceCountComputation`) try to use those enum values they have a multiplier value of 2 (rather than the expected 1).

Depending on what order the tests are run in, this may or may not cause a problem. The solution, I think, must be to move away from using an enum for `PeriodType` (and also avoid global mutable values). I can look into doing this over the next few days, but I think someone who is more familiar with the code base might be better for that job than me.

rivaldi8:
Good catch! That's indeed a problem. Fields in enums should be final.

I think the solution should be to move `PeriodType.mMultiplier` and its accessors to `Recurrence`. I'll look into it tomorrow, after some sleep.

Issue #598, opened 2016-10-10
Widget options (display name and whether to show balance)


czan:
I'm interested in adding some extra options to the widget when it's created. In particular I'm looking at adding:
 - an option to change the text that is displayed on the widget, so we're not forced to use the same string as the account name
 - an option to have a widget that does not display the current balance of the account, so we don't have to worry about people knowing our balances from our home screen

I have already written the code to do this. It's currently on top of the changes done to support books in `hotfix/patches` (which I assumed was the most up-to-date widget code), although I should be able to move it without too much trouble.

I thought it would be worth having a discussion about these options (and whether there are other options people would like to add) before I opened up a pull request, hence opening this issue.

codinguser:
Thanks for the code and for creating this issue so the features can be discussed.

Here are my thoughts on this:
1. What is the benefit of showing some text other than the account name on the widget? This seems like adding complexity with only marginal benefit. Or do you have some use-case in mind? 
@rivaldi8 what do you think?
2. Is the point of a widget with no amount displayed only just to act as a shortcut for opening the account and creating transactions? I see the utility in this. Or is there something else you have in mind?

These are changes which I would rather consider for the `develop` branch.

rivaldi8:
@codinguser I thought the same as you about 1. There should be a really good reason to justify the added complexity.

About 2, I'd like to add something. It would be nice to have the option to show the balance checked by default, unless the user has set a passcode, then it would be unchecked. This would align more with the privacy concerns of the user.

codinguser:
Yes, I like your point about 2. Unchecked by default if passcode is set 👍🏾






On Thu, Oct 13, 2016 at 5:37 PM +0200, "Àlex Magaz Graça" <notifications@github.com> wrote:












@codinguser I thought the same as you about 1. There should be a really good reason to justify the added complexity.



About 2, I'd like to add something. It would be nice to have the option to show the balance checked by default, unless the user has set a passcode, then it would be unchecked. This would align more with the privacy concerns of the user.

czan:
I have two main reasons for 1.

- Similarly to why I think 2 is a good idea, having to use the account name might leak information. I would like my accounts to have meaningful names when I am working with them, but when displayed on the home screen I may want to give it a less meaningful name in order to obscure its meaning to others. (In general, I would consider the widget "public information" and inside the application "private information", and I don't think the two should always be the same.)

[...]

PR #600, opened 2016-10-14
Make widgets able to hide the account balance

czan:
This commit adds an option to widgets which lets you hide the balance of the account that the widget is displaying. If you have set a passcode then the widget will default to not showing the account balance, otherwise it will default to showing the account balance.

This resolves part of #598.


rivaldi8:
Looks good to me and works as expected.

PR #601, opened 2016-10-15
Fix first run wizard crashes on screen rotation

### LINKED ISSUE #596: App crashes when changing the orientation from portrait to landscape in welcome section
1. Open the app for the first time
2. In welcome screen change the orientation of the phone from portrait to landscape

#### Expected behaviour
The app fits the orientation changes

#### Actual behaviour
The app crashes

#### Software specifications
* GnuCash Android version: 2.1.2-beta0
* System Android version: 5.0.2
* Device type: Motorola Moto G (second generation)

czan:
If the screen rotates while the first-run wizard is open then GnuCash will crash with a NullPointerException. To avoid this we need to make sure we create our WizardModel before it is used to try to restore our page.

This should fix #596.


rivaldi8:
Thanks for the pull request! As it's a small fix, would you mind sending it against hotfix/patches instead?

codinguser:
Thanks 👍

PR #602, opened 2016-10-19
Change widget button to open placeholder accounts

### LINKED ISSUE #592: Widget set on a placeholder account should list the subaccounts instead of a transaction form.
1. Create homescreen widget with placeholder account
2. Click the widget button to add transactions

#### Expected behaviour
3. It should have a button which just opens the account

#### Actual behaviour
4. It opens up to add a transaction to a placeholder account. 
Placeholder accounts cannot / should not contain transactions

#### Software specifications
* GnuCash Android version: all
* System Android version: all
* Device type: all

From uservoice feedback: https://gnucash.uservoice.com/forums/320493-gnucash-android/suggestions/16303108-widget-set-on-a-placeholder-account-should-list-th#

czan:
When a widget is created for a placeholder account we change the "add transaction" button on the widget to instead be an "information" icon, which opens up the account details when clicked.

Should fix #592.

### DIFF
--- a/app/src/main/res/layout/widget_4x1.xml
+++ b/app/src/main/res/layout/widget_4x1.xml
@@ -54,5 +54,14 @@
 	        android:src="@drawable/content_new_holo_light"
 	        android:background="@drawable/appwidget_bg_clickable"
 	        android:contentDescription="@string/description_add_transaction_icon"/>
+
+	   <ImageButton android:id="@+id/btn_view_account"
+		   android:layout_width="48dp"
+		   android:layout_height="wrap_content"
+		   android:layout_marginRight="@dimen/edge_padding"
+		   android:layout_marginLeft="@dimen/edge_padding"
+		   android:src="@drawable/action_about_holo_light"
+		   android:background="@drawable/appwidget_bg_clickable"
+		   android:contentDescription="@string/description_view_account_icon"/>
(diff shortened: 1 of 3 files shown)


czan:
I don't think I introduced the build failure. It looks like commit b5117fcb07ea80dca0c499933b75639831b3b196 failed in the same way.

codinguser:
@czan no, that was me :) 
sorry. Pushed changes to fix it already and then I'll trigger a rebuild of this PR

app/src/main/res/layout/widget_4x1.xml
```diff
@@ -54,5 +54,14 @@
 	        android:src="@drawable/content_new_holo_light"
 	        android:background="@drawable/appwidget_bg_clickable"
 	        android:contentDescription="@string/description_add_transaction_icon"/>
+
+	   <ImageButton android:id="@+id/btn_view_account"
```
codinguser: @czan What is the purpose of this button? Tapping the widget already opens the account does it not?

czan: Correct. I made it show an information icon primarily so the widget would look similar in either case. I originally had the button disappear, but it looked strange when a placeholder widget was next to a normal widget.

PR #605, opened 2016-10-25
Bugfix 604: Scheduled exports aren't executed if they're checked before the next run

### LINKED ISSUE #604: Scheduled exports aren't executed if they're checked before the next run
Scheduled exports won't be executed if the scheduled action service runs before the next run. The reason is that [ScheduledActionService.executeScheduledEvent](https://github.com/codinguser/gnucash-android/blob/d1c31033366ae7044e4b115b9ac021060660ef18/app/src/main/java/org/gnucash/android/service/ScheduledActionService.java#L145) sets the scheduled action's last run even if it hasn't actually been run. As the service runs twice a day, the period will never pass.
#### Software specifications
- GnuCash Android version: 2.1.3
- System Android version: 5.0
- Device type: bq Aquaris E4

rivaldi8:
I hope this time it's fixed for real. Please, check nothing else has slipped through.

Fixes #604

### DIFF
--- a/app/src/main/java/org/gnucash/android/service/ScheduledActionService.java
+++ b/app/src/main/java/org/gnucash/android/service/ScheduledActionService.java
@@ -127,7 +127,7 @@ public class ScheduledActionService extends IntentService {
      */
     private static void executeScheduledEvent(ScheduledAction scheduledAction, SQLiteDatabase db){
         Log.i(LOG_TAG, "Executing scheduled action: " + scheduledAction.toString());
-        int executionCount = scheduledAction.getExecutionCount();
+        int executionCount = 0;
 
         switch (scheduledAction.getActionType()){
             case TRANSACTION:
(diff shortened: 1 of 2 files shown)


rivaldi8:
Wait! don't merge it yet. I've just noticed that the transactions exported after the fix include the period since the scheduled export creation date instead of since the last export date. Not sure if it's something that should be fixed here, but I'd like to investigate first. I think there's an UserVoice ticket reporting a similar issue.

rivaldi8:
It's ok to merge this pull request. I'll verify the other problem with some testing and I'll open another bug report for it.

codinguser:
I'm planning to remove scheduled exports/backups to their own table and have their own service in v2.2.0 anyway. So I would advise to invest time only for critical stuff which affects users. 
Separating the scheduled backups and transactions will mean we have to redo most of this logic anyway, hopefully then in a more stable way.

codinguser:
@rivaldi8 I got some reports of all transactions always being exported. That is one thing we should definitely fix ASAP since it affects user workflow. Please let me know if you find a cure :)

rivaldi8:
Ok, I was working on these export-related bugs because I was affected by them :) I've also seen the one about all transactions being exported. I was about to work on it after fixing #609, which I thought it was related although I'm no longer sure.

PR #606, opened 2016-10-27
Make widgets backwards compatible

czan:
If we can't find the new "hide account balance" setting, we should default to the old behaviour: show the account balance.


rivaldi8:
Thanks for the fix!

Issue #517, opened 2016-07-11
Feature request: Enable/Disable inclusion of future transactions in total


elmerehbi:

codinguser:
Duplicate of #465

PR #612, opened 2016-11-10
Bugfix 571: Multi-currency transactions are ignored by QIF

### LINKED ISSUE #571: Multi-currency transactions are ignored by QIF
#### Expected behaviour

All transactions should be included in the file, even if they are only using the converted amount
#### Actual behaviour

Multicurrency transactions are completely ignored
1. Create multi-currency transactions
2. Export QIF from the app
#### Software specifications
- GnuCash Android version: 2.1.1
- Android version: Any
- Device type: Any

rivaldi8:
Should fix #571. It's been too easy, so I'm not really sure :)


codinguser:
Simple is good ;)
And normal QIF exports are not affected by this right?
Also, as per your example in #517, the correct amount is read from the split quantity when exporting the Foreign account right?

rivaldi8:
Sorry, I'm late.

I guess you already checked, but just in case I checked again. Normal QIF don't seem to be affected. The splits are exported with the converted amount for the Foreign account, which I understand it's what should happen.

Well, let's see what happens with real-world usage :)

PR #619, opened 2016-11-25
Bugfix 617: Weekly scheduled exports are listed with the wrong day

### LINKED ISSUE #617: Weekly scheduled exports are listed with the wrong day
Create a weekly scheduled export on Sundays.

#### Expected behaviour
The scheduled export is listed as "Weekly on Sunday".

#### Actual behaviour
The scheduled export is listed as "Weekly on Thursday".

Reported through UserVoice (ticket 143).

rivaldi8:
Fixes #617 

There some are important changes, so I thought you might want to review it.

Instead of storing the days of the week as a string ("MO;TH") in `mByDay`, which is painful to work with, I've changed it to store as a List of Calendar constants (`Calendar.MONDAY`) instead. If later we want to support specific days in the month, we can just change the list type to some other class to hold the new information.

As this was meant for hotfix/patches, I haven't changed the database structure but, in my opinion, we should think of storing `recurrence_byday` in a separate table. Apart from not conforming with the first normal form, the parsing of the column (`RecurrenceDbAdapter.{byDaysToString,stringToByDays}` ) makes it awkward to work with.

### DIFF
--- a/app/src/main/java/org/gnucash/android/db/adapter/RecurrenceDbAdapter.java
+++ b/app/src/main/java/org/gnucash/android/db/adapter/RecurrenceDbAdapter.java
@@ -20,12 +20,17 @@ import android.database.Cursor;
 import android.database.sqlite.SQLiteDatabase;
 import android.database.sqlite.SQLiteStatement;
 import android.support.annotation.NonNull;
+import android.support.annotation.Nullable;
 
 import org.gnucash.android.app.GnuCashApplication;
 import org.gnucash.android.model.PeriodType;
 import org.gnucash.android.model.Recurrence;
 
 import java.sql.Timestamp;
+import java.util.ArrayList;
+import java.util.Calendar;
+import java.util.Collections;
+import java.util.List;
 
 import static org.gnucash.android.db.DatabaseSchema.RecurrenceEntry;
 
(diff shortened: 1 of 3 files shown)


codinguser:
Thanks for the pr. The recurrence parsing looks much cleaner now. We should merge this. 

It fixes #617 on a UI level, but I think the bug in the recurrence functionality will still exist. By this I mean that the day will still be ignored in the actual recurrence schedule. We can fix this by setting the start time of the `Recurrence` to the next occurrence of the day set in `byday`. This combined with a weekly period will mean that the schedule always happens on the same day set by the user. 

When there are multiple days per week in the recurrence, this is also ignored at the moment. Ideally I would prefer to turn this option off in the picker, but I didn't find a way. There are two options to handling these kinds of recurrences properly: 

1. Create multiple recurrences db entries with the same schedule but each one starting on the different day (from `byday`) -- a.k.a simple but inelegant solution
2. Modify the ScheduledActionService to properly handle weekly recurrences with multiple days per week -- a.k.a more elegant and also more complex.

What do you think? 

In the meantime, if you have no objections I will merge this pr.

rivaldi8:
Yes, to keep it simple, this fixes just the UI issue. I think it's better to deal with the actual running of the scheduled exports in a separate issue.

I'm not sure I understand what you propose about setting the start date of the `Recurrence`. `ScheduledActionService` checks if it should run an export (indirectly) from the last run date. Anyway, I think it would be better if we defined first which schedule features we want to support and how we deal with them when syncing to and from GnuCash desktop. This would it easier to make decisions. We can discuss it in the mailing list or in another issue.

No objections to merging :)

PR #622, opened 2016-12-04
Sort the transfer accounts by favorite before sorting them alphabetically

davidlandry93:
This is a series of commits that puts the favorite accounts on top of the other accounts transfer accounts list. It makes it easier to add transactions if you use some transfer accounts more frequently.

I was annoyed at the app because I only really use two transfer accounts and I had to scroll through all my accouts to get to my credit card.

To avoid any confusion, I made it so that the favorite "star" would show beside the name of the transfer accounts that are favorite. That way the user knows why they appear first.

Thanks for the consideration!


app/src/main/java/org/gnucash/android/util/FavoritableQualifiedAccountNameCursorAdapter.java
```diff
@@ -0,0 +1,83 @@
...
+     * Binds an account name and its favorite status to an appropriate view.
+     * @param view The view to bind the account to
+     * @param context Application context
+     * @param cursor Cursor to accounts
+     */
+    @Override
+    public void bindView(View view, Context context, Cursor cursor) {
+        super.bindView(view, context, cursor);
+
+        Integer is_favorite = cursor.getInt(cursor.getColumnIndex(DatabaseSchema.AccountEntry.COLUMN_FAVORITE));
```
rivaldi8: Please, try to follow Java coding conventions. Variables should use camel case. For example, `isFavorite` instead of `is_favorite`.

app/src/main/res/layout/favorite_transfer_account_spinner_item.xml
```diff
@@ -0,0 +1,24 @@
+<?xml version="1.0" encoding="utf-8"?>
+<LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
```
rivaldi8: Please, try to fix the editor warnings if possible. For example, in this line it says the `<LinearLayout>` and its contents can be replaced by a single `<TextView>` with a `drawableEnd` attribute.

rivaldi8:
Thanks for the pull request! I think it's a good idea.

I've checked all the uses of `QualifiedAccountNameCursorAdapter` and I think all of them would be improved if the favorite accounts were listed first and with the star. So instead of implementing it as a separate `FavoritableQualifiedAccountNameCursorAdapter` class, just add the feature to `QualifiedAccountNameCursorAdapter`.

There's one thing I don't like, though. I agree that the favorite stars help to avoid confusion about why those entries appear first when showing the list. However, when a favorite account is selected, the star is still shown. I think it stands out too much and adds clutter. Also it looked a bit confusing when I first saw it. Could you try to hide it when the spinner list closed. It may be possible to do so from [AdapterView.OnItemSelectedListener.onItemSelected](https://developer.android.com/reference/android/widget/AdapterView.OnItemSelectedListener.html?hl=es-419#onItemSelected(android.widget.AdapterView%3C?%3E,%20android.view.View,%20int,%20long)). If it's not possible, I'd rather not show it for now.



Please, let me know if you need any help with this.

davidlandry93:
I think you are right, I'll look into it this weekend probably. Thanks!

rivaldi8:
Thanks! Merged.

PR #624, opened 2016-12-07
Logs

pnemonic78:


rivaldi8:
Sorry, I can't merge this pull request. There are several problems with it:
- Pull requests shouldn't be sent to 'master' branch (See [CONTRIBUTING.md])(https://github.com/codinguser/gnucash-android/blob/master/.github/CONTRIBUTING.md)
- It mixes unrelated changes. Code refactors in the first commits with changes in the definition of log tags in the later ones.
- Changes like the ones in the first commits should be avoided. It's better to do so in the way to fixing some bug or adding some feature.
- Commit messages should describe the change properly, be atomic... See, [these guidelines](http://who-t.blogspot.com.es/2009/12/on-commit-messages.html) for example.

Apart from this, it's better to start contributing to a project by working on some small issue from the bug tracker or that you might have found.

PR #628, opened 2016-12-13
Bugfix 603: scheduled transactions with multiple splits end up going to imbalance

pnemonic78:
Template transactions have splits with zero quantities but non-zero values.
The quantity's currency might be different than the value's currency.

### DIFF
--- a/app/src/main/java/org/gnucash/android/importer/GncXmlHandler.java
+++ b/app/src/main/java/org/gnucash/android/importer/GncXmlHandler.java
@@ -688,14 +688,14 @@ public class GncXmlHandler extends DefaultHandler {
                 break;
             case GncXmlHelper.TAG_TRANSACTION:
                 mTransaction.setTemplate(mInTemplates);
-                Split imbSplit = mTransaction.createAutoBalanceSplit();
-                if (imbSplit != null) {
-                    mAutoBalanceSplits.add(imbSplit);
-                }
                 if (mInTemplates){
                     if (!mIgnoreTemplateTransaction)
                         mTemplateTransactions.add(mTransaction);
                 } else {
+                    Split imbSplit = mTransaction.createAutoBalanceSplit();
+                    if (imbSplit != null) {
+                        mAutoBalanceSplits.add(imbSplit);
+                    }
                     mTransactionList.add(mTransaction);
                 }
                 if (mRecurrencePeriod > 0) { //if we find an old format recurrence period, parse it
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/gnucash/android/model/Transaction.java
```diff
@@ -268,7 +268,11 @@ public Money getBalance(String accountUID){
     public Money getImbalance(){
         Money imbalance = Money.createZeroInstance(mCurrencyCode);
         for (Split split : mSplitList) {
-            if (!split.getQuantity().getCommodity().getCurrencyCode().equals(mCurrencyCode)) {
+            Money quantity = split.getQuantity();
+            if (quantity.isAmountZero()) {
```
rivaldi8: I'd say `quantity` should never be zero (see [this comment](https://github.com/codinguser/gnucash-android/issues/506#issuecomment-231603278)). Am I missing something?

app/src/main/java/org/gnucash/android/importer/GncXmlHandler.java
```diff
@@ -688,14 +688,14 @@ public void endElement(String uri, String localName, String qualifiedName) throw
...
-                    mAutoBalanceSplits.add(imbSplit);
-                }
                 if (mInTemplates){
                     if (!mIgnoreTemplateTransaction)
                         mTemplateTransactions.add(mTransaction);
                 } else {
+                    Split imbSplit = mTransaction.createAutoBalanceSplit();
+                    if (imbSplit != null) {
+                        mAutoBalanceSplits.add(imbSplit);
+                    }
```
rivaldi8: If we get a transaction with unbalanced splits (scheduled or normal), I'd say it's either we've failed to import it correctly or there's something wrong in the file.

app/src/main/java/org/gnucash/android/model/Transaction.java
pnemonic78: missing the fact that template transactions have splits with zero quantity, but non-zero values.

rivaldi8: In which case would we have a valid template transaction with zero quantity and non-zero value? It doesn't seem possible from the Android application. How do you create such template transaction?

pnemonic78: it happens when importing from Desktop.

rivaldi8: Then it's an import problem that should be fixed instead of worked around.

rivaldi8:
Sorry, I can merge these changes. They are working around the problem instead of fixing it.

PR #629, opened 2016-12-13
Bugfix 564: Pressing "Enable Google Drive" checkbox option leaves it unchecked

pnemonic78:
Google API connection is static, meaning that the settings fragment is not in scope for its UI.

### DIFF
--- a/app/src/main/java/org/gnucash/android/ui/settings/BackupPreferenceFragment.java
+++ b/app/src/main/java/org/gnucash/android/ui/settings/BackupPreferenceFragment.java
@@ -70,7 +70,7 @@ import java.util.List;
 public class BackupPreferenceFragment extends PreferenceFragmentCompat implements
 		Preference.OnPreferenceClickListener,
 		Preference.OnPreferenceChangeListener,
-		OnCloudStorageListener {
+		OnGoogleDriveListener {
 
 	/**
 	 * Collects references to the UI elements and binds click listeners
@@ -276,7 +276,6 @@ public class BackupPreferenceFragment extends PreferenceFragmentCompat implement
 		if (appFolderId != null){
 			sharedPreferences.edit().remove(getString(R.string.key_google_drive_app_folder_id)).commit(); //commit (not apply) because we need it to be saved *now*
 			mGoogleApiClient.disconnect();
-			onGoogleDriveFolderForgot(appFolderId);
 		} else {
 			mGoogleApiClient.connect();
 		}
@@ -300,8 +299,8 @@ public class BackupPreferenceFragment extends PreferenceFragmentCompat implement
 		return getGoogleApiClient(context, null);
 	}
 
-	public static GoogleApiClient getGoogleApiClient(final Context context, OnCloudStorageListener listener) {
-		final WeakReference<OnCloudStorageListener> weakListener = new WeakReference<>(listener);
+	public static GoogleApiClient getGoogleApiClient(final Context context, OnGoogleDriveListener listener) {
+		final WeakReference<OnGoogleDriveListener> weakListener = new WeakReference<>(listener);
 		mGoogleApiClient = new GoogleApiClient.Builder(context)
 				.addApi(Drive.API)
 				.addScope(Drive.SCOPE_APPFOLDER)
@@ -328,7 +327,7 @@ public class BackupPreferenceFragment extends PreferenceFragmentCompat implement
 									sharedPreferences
 											.edit().putString(context.getString(R.string.key_google_drive_app_folder_id),
 											folderId).commit(); //commit because we need it to be saved *now*
-									OnCloudStorageListener listener = weakListener.get();
+									OnGoogleDriveListener listener = weakListener.get();
 									if (listener != null) {
 										listener.onGoogleDriveFolderCreated(folderId);
 									}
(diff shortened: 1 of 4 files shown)


app/src/main/java/org/gnucash/android/ui/settings/BackupPreferenceFragment.java
```diff
@@ -422,12 +437,32 @@ public void onActivityResult(int requestCode, int resultCode, Intent data) {
...
 						break;
-					toggleDropboxPreference(pref);
+					toggleGoogleDrivePreference(pref);
 				}
 				break;
 		}
 	}
+
+	@Override
+	public void onGoogleDriveFolderCreated(String folderId) {
```
rivaldi8: The `folderId` parameter isn't used, remove it.

app/src/main/java/org/gnucash/android/ui/settings/BackupPreferenceFragment.java
```diff
@@ -273,6 +276,7 @@ private void toggleGoogleDriveSync(){
 		if (appFolderId != null){
 			sharedPreferences.edit().remove(getString(R.string.key_google_drive_app_folder_id)).commit(); //commit (not apply) because we need it to be saved *now*
 			mGoogleApiClient.disconnect();
+			onGoogleDriveFolderForgot(appFolderId);
```
rivaldi8: Why the `onGoogleDriveFolderForgot` method? Why not just inline the code here?

app/src/main/java/org/gnucash/android/ui/settings/BackupPreferenceFragment.java
```diff
@@ -292,64 +296,75 @@ private void toggleOwnCloudSync(Preference pref){
...
 	}
 
-
 	public static GoogleApiClient getGoogleApiClient(final Context context) {
-		return new GoogleApiClient.Builder(context)
+		return getGoogleApiClient(context, null);
+	}
+
+	public static GoogleApiClient getGoogleApiClient(final Context context, OnCloudStorageListener listener) {
+		final WeakReference<OnCloudStorageListener> weakListener = new WeakReference<>(listener);
```
rivaldi8: I'm not sure I see the need for the `WeakReference`. `listener` is only set to a non-null value when called internally from `onCreate()`. So if the Fragment is destroyed, the `GoogleApiClient` should be garbage collected, shouldn't it?

app/src/main/java/org/gnucash/android/ui/settings/OnCloudStorageListener.java
```diff
@@ -0,0 +1,38 @@
...
+ */
+
+package org.gnucash.android.ui.settings;
+
+/**
+ * Listener interface for cloud storage events.
+ *
+ * @author moshe.w
+ */
+public interface OnCloudStorageListener {
```
pnemonic78: I intended the listener to be generic that it could also be used for other cloud storages such as OneDrive, etc.

app/src/main/java/org/gnucash/android/ui/settings/BackupPreferenceFragment.java
pnemonic78: Intention to notify listener(s) that the folder is no longer available and that they should do something about it.

app/src/main/java/org/gnucash/android/ui/settings/BackupPreferenceFragment.java
pnemonic78: the problem is that the function is "static". any references to heavy objects like fragments or activities will cause memory leaks.

app/src/main/java/org/gnucash/android/ui/settings/BackupPreferenceFragment.java
pnemonic78: The parameter isn't used now, but the listener interface provides it for those listener(s) that would need it. Other examples of such methods are ListView adapter events that you don't always use all the parameters.

app/src/main/java/org/gnucash/android/ui/settings/OnCloudStorageListener.java
rivaldi8: Ok, then just remove the `public` access as suggested by the code inspector. It can be added later, if really needed.

app/src/main/java/org/gnucash/android/ui/settings/BackupPreferenceFragment.java
rivaldi8: Ok. Shouldn't it be `on...Forgotten`?

app/src/main/java/org/gnucash/android/ui/settings/BackupPreferenceFragment.java
rivaldi8: Please, if it's not used, remove it. Now it just adds noise. It can be easily added later, if it's really needed.

app/src/main/java/org/gnucash/android/ui/settings/OnCloudStorageListener.java
rivaldi8: I forgot that, if it's to be generic, you should make the methods names generic too.

codinguser:
@pnemonic78 thanks for the pull request. 

In addition to the issues raised by @rivaldi8 , I have been reviewing this PR since yesterday and I am disinclined to merge it. It is hard for me to figure out what exactly the PR does. It says to fix the checkbox for Google Drive, but at the same time it seems to be trying to abstract other concepts and not going all the way. 

I would prefer a much simpler and direct fix for the GDrive checkbox in the Settings if that was the goal. Otherwise, for bigger changes, it is usually better to create an issue where it can be discussed and you get valuable feedback before coding it. Then send the PR when the change and all accompanying refactoring is complete.

Also, in the future, it would be help if you give a clearer description together with the pull request (esp for non-trivial fixes). That helps to be able to more effectively review the code. If there is another issue where it is already discussed, you could just link to that. 

Again, thanks for your work. I know/appreciate the work that goes into creating one. So do not be discouraged, but keep the code coming!

codinguser:
There is no checkbox for Google Drive anymore. 
So I'm closing this.

PR #639, opened 2017-01-15
proper handling of custom date range selections - fixes #611

### LINKED ISSUE #611: Custom date range doesn't work properly in reports
From user email: 
I found a small bug where when viewing charts using custom range period, the chart always display the data until today.

So the beginning date is correct, but the data shown always include everything until today, and not until the specified end date.

HeinrichLohse:
I am using this app for a month now and was bothered that the date range was not properly set when I was trying to enjoy the pie chart report. This small fix does the trick for me. Since I have only little experience with git and I am completely new to github I want to apologize in advance if I did something wrong with the workflow.


codinguser:
Thanks for the fix @HeinrichLohse 
Could you please send this pull request against the branch `hotfix/patches`? 
We never merge directly into `master` branch!

Issue #633, opened 2016-12-31
App crashes when changing orientation while viewing reports


diogocp:
#### Steps to reproduce the behaviour
1. Open one of the reports (pie chart, line chart, bar chart or sheet).
2. Turn the phone to change the screen orientation.

#### Expected behaviour
Report shows in the new screen orientation.

#### Actual behaviour
App crashes.

#### Software specifications
* GnuCash Android version: 2.1.3
* System Android version: Android 6.0.1, CyanogenMod 13.0
* Device type: Nexus 4

rivaldi8:
Confirmed.

PR #645, opened 2017-01-23
Preserve report type accross rotation

GianpaMX:
Not sure if this is the best way to avoid the crash but for an unknown reason the `onAttachFragment()` is being called from parent's `onCreate()`

* fixes codinguser/gnucash-android#633


codinguser:
Thanks for the PR @GianpaMX 
So I'm unsure from what you wrote: Does this fix the crashing or not?
Also, could you send the PR against the `develop` branch. We never merge anything directly into `master`

GianpaMX:
Yes this fixes the crashing. I'm closing this PR and opening a new one against `develop`; just code review it before merging to make sure the solution makes sense

PR #648, opened 2017-01-24
Currency-commodity migration changes

rivaldi8:
I was a bit wary about the second commit (86f2343), so I'd feel better if you confirmed it's ok.

### DIFF
--- a/app/src/main/java/org/gnucash/android/db/adapter/AccountsDbAdapter.java
+++ b/app/src/main/java/org/gnucash/android/db/adapter/AccountsDbAdapter.java
@@ -1176,7 +1176,7 @@ public class AccountsDbAdapter extends DatabaseAdapter<Account> {
 
                 Transaction transaction = new Transaction(GnuCashApplication.getAppContext().getString(R.string.account_name_opening_balances));
                 transaction.setNote(getAccountName(accountUID));
-                transaction.setCurrencyCode(currencyCode);
+                transaction.setCommodity(Commodity.getInstance(currencyCode));
                 TransactionType transactionType = Transaction.getTypeForBalance(getAccountType(accountUID),
                         balance.isNegative());
                 Split split = new Split(balance.abs(), accountUID);
--- a/app/src/main/java/org/gnucash/android/importer/GncXmlHandler.java
+++ b/app/src/main/java/org/gnucash/android/importer/GncXmlHandler.java
@@ -438,8 +438,8 @@ public class GncXmlHandler extends DefaultHandler {
                 break;
             case GncXmlHelper.TAG_COMMODITY_ID:
                 String currencyCode = mISO4217Currency ? characterString : NO_CURRENCY_CODE;
+                Commodity commodity = mCommoditiesDbAdapter.getCommodity(currencyCode);
                 if (mAccount != null) {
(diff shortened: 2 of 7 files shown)


app/src/main/java/org/gnucash/android/db/adapter/AccountsDbAdapter.java
```diff
@@ -1176,7 +1176,7 @@ public boolean isFavoriteAccount(String accountUID){
 
                 Transaction transaction = new Transaction(GnuCashApplication.getAppContext().getString(R.string.account_name_opening_balances));
                 transaction.setNote(getAccountName(accountUID));
-                transaction.setCurrencyCode(currencyCode);
```
codinguser: The currency_code column in the transactions database might be a non-null column. Could you check?
In that case, we would need to set it to some default string or for the time-being. 
During the big db migration coming up, I'd get rid of that column.

app/src/main/java/org/gnucash/android/importer/GncXmlHandler.java
```diff
@@ -675,7 +675,7 @@ public void endElement(String uri, String localName, String qualifiedName) throw
                     //the split amount uses the account currency
                     mSplit.setQuantity(new Money(mQuantity, getCommodityForAccount(characterString)));
                     //the split value uses the transaction currency
-                    mSplit.setValue(new Money(mValue, mCommoditiesDbAdapter.getCommodity(mTransaction.getCurrency().getCurrencyCode())));
+                    mSplit.setValue(new Money(mValue, mTransaction.getCommodity()));
```
codinguser: Looks so much better 😆

app/src/main/java/org/gnucash/android/db/adapter/AccountsDbAdapter.java
rivaldi8: I'm not sure I follow you here. The `transactions.currency_code` column is still being set with the currency code from the Commodity in Transaction. So, for now, it shouldn't be a problem.

We should probably get rid of the column with the db migration, but I think it's used in some view. So it might not be as easy as it seems.

app/src/main/java/org/gnucash/android/importer/GncXmlHandler.java
rivaldi8: Improving little by little :-)

rivaldi8:
I dealt with the `TransactionRecorder` [here](https://github.com/codinguser/gnucash-android/pull/648/files#diff-e31eb15ec2f9f74d1679f8e914c50914). The other references are not important. I just want to get rid of `Currency` for now. There will be more commits to finish migrating from Currency to Commodity. I'm not sure if you misunderstood this pull request as doing it all :)

By the way, when is `TransactionRecorder` used? It mentions `Transaction.createIntent` but there isn't any use.

About UI tests, It's been impossible to run them with my phone. However, I've just got lent a phone where they work. So I'm running the tests again.

app/src/main/java/org/gnucash/android/db/adapter/AccountsDbAdapter.java
codinguser: oh I see, my bad. 
Then its all good.

PR #650, opened 2017-01-27
List favorite accounts first in Recents tab

rivaldi8:
**Not to be merged**

While using the application, I usually have the _Favorites_ tab selected. There are the accounts I use most of the time (of course). However, there are also some accounts that I use less frequently. Then, the _Recents_ tab come in handy. The problem is, if later I need a favorite account, I might have to scroll down or switch back to _Favorites_.

What I've been thinking for some time is to list favorite accounts first in _Recents_. That would make easier to access both kinds of accounts and _Favorites_ tab could be removed.

What do you think? If you like the idea I can finish implementing it.


codinguser:
I like the idea. When I think of it, it makes sense that the favourite accounts are probably also the ones you use most often and so they would be in the recents list anyway. 
This way, we can just show the favourites first, and then the other recent accounts. Two lists instead of 3. Nice. 

But I would like it if we could do it as a sectioned list. Like a list section header which says Favourites and then when you finish scrolling the favourites, you get the next section "Recently Used" accounts.

rivaldi8:
I like the headers idea. I'll work on it then.

PR #671, opened 2017-04-13
Start autocompleting transactions with one character

rivaldi8:
Less typing is better :)

I guess you just left the default. I can't find any reason for having to type two characters instead of one but, just in case, I leave it for you to approve.


codinguser:
Yes, I just left the default in there, but it was conscious decision - db hits (perf perspective), recurring transactions with same starting letting in the description etc. 
I guess we could change it to just 1 letter now and see how that works, especially with large books.

rivaldi8:
In my phone, with around 800 transactions it's almost instantaneous. Maybe with a large database and an old phone the lag might be noticeable, but I think in most cases it will be an improvement. It's just one line, so it can be reverted easily.

PR #680, opened 2017-04-25
Bugfix 641: Weekday is ignored in scheduled exports

rivaldi8:
Am I late for 2.2?

Please, check I haven't missed any corner case :)

### DIFF
--- a/app/src/main/java/org/gnucash/android/model/ScheduledAction.java
+++ b/app/src/main/java/org/gnucash/android/model/ScheduledAction.java
@@ -24,6 +24,7 @@ import org.joda.time.LocalDateTime;
 
 import java.sql.Timestamp;
 import java.text.SimpleDateFormat;
+import java.util.Calendar;
 import java.util.Date;
 import java.util.Locale;
 import java.util.TimeZone;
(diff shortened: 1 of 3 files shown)


codinguser:
No, it's not late. I just wanted to get a first beta out as soon as
possible so we can get actual user feedback.

I'll check it when I get back to my desk. Thanks!

rivaldi8:
Sorry, I forgot to switch back to the full unit test suite before sending the pull request. It seems the tests have to be updated to set `byDays` for weekly actions.

Today I have a busy day. I'll fix it tomorrow.

rivaldi8:
It should be fixed now.

codinguser:
Cool, thanks. So `byday` is always set by the recurrence dialog, right?

rivaldi8:
Yes, the dialog forces you to set at least on weekday. However, after a more thorough inspection, I've found `GncXmlHandler` never sets the recurrence's `byDays`. It doesn't parse the weekdays.

Until we don't implement the parsing for imported weekly transaction, I'd add at least one weekday to `byDays`. We could just use the weekday of the start date. That way it would be executed at least once per week. In fact I'd say that's how it worked until now.

Also, although with a warning, GnuCash desktop allows you to define a weekly transaction without setting any weekday. In that case, it's never executed. So, I'll have to add a check for empty `byDays` too.

codinguser:
@rivaldi8 sounds like a plan 👍

PR #684, opened 2017-05-06
Modify KRW fraction of ISO4217 currency settings.

hkjinlee:
Chon(=Jeon), the smaller part of KRW, is no longer used in Korea.
Leaving it as it is causes annoying .00 attached after every value.
Better change currency settings.

Reference:
- https://en.wikipedia.org/wiki/ISO_4217 : 
Number of digits after decimal separator is 0 for KRW
- https://en.wikipedia.org/wiki/South_Korean_won :
“The jeon is no longer used for everyday transactions”


hkjinlee:
I posted the same PR to the Gnucash desktop, and the discussion is going on there. Will update the comment when the discussion yields some viable results.

https://github.com/Gnucash/gnucash/pull/146

codinguser:
Relevant thread is here: https://bugzilla.gnome.org/show_bug.cgi?id=666536

If gnucash maint decides to update it, so will we

codinguser:
There are also considerations about how to handle existing accounts, so we have to be cautious.

codinguser:
Oh, I just noticed that this PR is issued against `master` branch. 
We never merge into master, so I will close this now. But when there is a decision, we can act on it in `develop` branch.

PR #700, opened 2017-06-26
Fix bug #699

### LINKED ISSUE #699: Transaction details screen footer disappears
1. Click on existing transaction
1. Click on edit transaction FAB
1. Click on splits icon
1. Add a split
1. Click "save"
1. Click "save" again

#### Expected behaviour
Footer visible.

#### Actual behaviour
Footer removed (balance and date).

#### Software specifications
* GnuCash Android version: 2.2.0
* System Android version: 6.0
* Device type: Any

pnemonic78:
Adding a split was removing too many rows from the details table.

Fixes #699


app/src/main/java/org/gnucash/android/ui/transaction/TransactionDetailActivity.java
```diff
@@ -175,8 +175,15 @@ private void refresh(){
...
-        mDetailTableLayout.removeViews(0, (int)splitCount);
+        final int childCount = mDetailTableLayout.getChildCount();
+        for (int i = 0; i < childCount; i++) {
+            if (mDetailTableLayout.getChildAt(0).getId() == R.id.row_split_amount_info) {
+                mDetailTableLayout.removeViewAt(0);
+            } else {
+                // All splits are at the top of the table.
+                break;
+            }
+        }
```
rivaldi8: It works, but I'd like a simpler solution.

We know there will always be at least 4 rows in `fragment_transaction_details` (`mDetailTableLayout`) apart from the splits: `row_balance`, `row_time_and_date`, `row_trn_notes` and `row_trn_recurrence`. Then, could we substitute the for loop with this?
```
mDetailTableLayout.removeViews(0, childCount - 4);
```

pnemonic78: Done.

rivaldi8: Sorry, I must be missing something, but I can't see it. Why have you used a for loop instead of `removeViews`?

pnemonic78:
Replaced for-loop.

rivaldi8:
Thanks! Merged manually onto hotfix/patches.

Issue #343, opened 2015-06-16
Export notes with QIF


codinguser:

hargi:
Not exporting notes in QIF and/or XML currently is a no-go for me.
Would love to see that implemented.

codinguser:
@hargi notes are exported in XML. 
QIF is the tricky one

rivaldi8:
Which QIF field should be used?

MisterY:
Isn't M used for notes/description? And it is already utilised.

rivaldi8:
That's right. M is used for the description of transactions. We need another field for notes.

MisterY:
Ah, I see, you are using the terminology from GnuCash, which does not match QIF 100%.
Unfortunately, with QIF, description = notes. 
What is missing here is the Payee.

Which brings up an interesting point. I'm also very interested in adjusting the QIF export (starting with #690) and it might be useful to reshuffle the current field values. The description could be exported as the Payee (P) field, and the notes would go as the notes/description (M) field.

This would also solve my issues as I'm actually currently exporting to Quicken. Unfortunately, over the years, I was unable to transition to desktop GnuCash due to the lack of investment features but that's another story.

In any case, I'm happy to contribute to these changes (I'm also actively contributing to Money Manager Ex for Android) in QIF export & other features, if there's a concensus.

codinguser:
Thanks for the input @MisterY 
But isn't Payee `a person to whom money is paid or is to be paid, especially the person to whom a check is made payable`. 
Would that not represent a conflict if in the future the app actually supports a Payee?

MisterY:
Hi, @codinguser . While I can't comment on how the app will develop in the future and whether it will include additional fields, I am fairly certain that QIF will not include that field. So, here I was strictly speaking about QIF export and fields, and what to put in there. I, for example, am using Quicken and would like to import the transactions there.

Besides, I see no point in adding another field for the other party in a transaction when the amount already identifies in which direction the money flows.

codinguser:
@MisterY I see what you mean. 
So my question would be, does GnuCash desktop import the QIF description correctly when the description is set to the Payee field. If it works there, then we can implement it. Because our goal is to maximise compatibility with GnuCash desktop.

MisterY:
Yes, this is something to be confirmed. Having both Description and Notes fields makes Payee unnecessary. Either that or one of the other fields​ would be redundant.
[...]

PR #706, opened 2017-07-03
Adding Payee/Description field to QIF export.

MisterY:
Adding Payee/Description field to QIF export and moving the Memo to Memo QIF field.
Implements #343 as discussed there.

### DIFF
--- a/app/src/test/java/org/gnucash/android/test/unit/export/QifExporterTest.java
+++ b/app/src/test/java/org/gnucash/android/test/unit/export/QifExporterTest.java
@@ -41,7 +41,12 @@ import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.robolectric.annotation.Config;
 
+import java.io.BufferedReader;
 import java.io.File;
+import java.io.FileNotFoundException;
+import java.io.FileReader;
+import java.io.IOException;
+import java.io.InputStreamReader;
 import java.util.List;
 
 import static org.assertj.core.api.Assertions.assertThat;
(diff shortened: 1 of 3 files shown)


rivaldi8:
Merged manually onto hotfix/patches (26277c6). Thanks!

app/src/test/java/org/gnucash/android/test/unit/export/QifExporterTest.java
```diff
@@ -144,4 +149,46 @@ public void multiCurrencyTransactions_shouldResultInMultipleQifFiles(){
         assertThat(file.length()).isGreaterThan(0L);
     }
 
+    //@Test
+    public void description_and_memo_field_test() {
```
rivaldi8: Please, follow Java naming conventions next time. `testDescriptionAndMemoField()` would've been better.

MisterY: Alright, thanks for the suggestion. I do, however, disagree that camel-case is an appropriate syntax for long test names. It is completely unreadable for anything longer than three words.
Even the name of the existing test combines the two standards, which is even worse, in my opinion.

rivaldi8: Yes, I also disagree with some conventions, but not following them just makes the code stand out unnecessarily.

The tests combining the two styles is also a convention :-) I can't find the reference now, but more or less it follows the format: `someAction_whatDoYouExpectFromIt()`.

PR #711, opened 2017-07-31
edited gradle.properties to disable build cache

wanching0730:
I found one issue when I was running an automated test in Jenkins. The issue is caused by the new feature of build-cache. It is using the default build-cache folder which is a long path. Hence, I added one line of code into the gradle.properties in order to disable build-cache. Hope this will help. Thank you


codinguser:
Hi, thanks for the pull request. 
Does disabling the build cache not make the build cache not make the build take even longer? We wouldn't want that. 
Also, we do not accept pull requests on master. Please see our CONTRIBUTION guidelines here: https://github.com/codinguser/gnucash-android/blob/master/.github/CONTRIBUTING.md

codinguser:
See comments on #713

PR #713, opened 2017-08-01
Made few minor changes

wanching0730:
Made some changes on few minor issues. 
1. Fixed typo
2. Deleted unused resource
3. Change to "ifRoom" instead of "always" for app:showAsAction

Thank you.


codinguser:
Hi @wanching0730 
as discussed in the previous pull request, we do not want to disable build caching. This actually helps speed up the builds which already take considerable time. [Caching of pre-dexed remote dependencies is already disabled because the app is multidex](https://developer.android.com/studio/build/build-cache.html).

Furthermore, when submitting pull requests, make sure that the PR fixes only one issue. It makes it easier to review. This PR edits whitespace, changes the bundle gradle, adds binary dumps among other things. It cannot be merged in this state under any circumstance, sorry.

PR #714, opened 2017-08-02
Zip multiple exported QIF files into one for all export destinations

### LINKED ISSUE #696: Zip multiple exported QIF files into one for all export destinations
1. Export transactions with multiple currencies as QIF to ownCloud
2. View files at destination

#### Expected behaviour
3. One zip file containing all exported files

#### Actual behaviour
3. One QIF per currency at destination

#### Software specifications
* GnuCash Android version: all
* System Android version: all
* Device type: all

rivaldi8:
Fixes #696

### DIFF
--- a/app/src/main/java/org/gnucash/android/export/ExportAsyncTask.java
+++ b/app/src/main/java/org/gnucash/android/export/ExportAsyncTask.java
@@ -283,25 +283,10 @@ public class ExportAsyncTask extends AsyncTask<ExportParams, Void, Boolean> {
         if (mExportedFiles.size() > 0){
             try {
                 OutputStream outputStream = mContext.getContentResolver().openOutputStream(exportUri);
-                ZipOutputStream zipOutputStream = new ZipOutputStream(outputStream);
-                byte[] buffer = new byte[1024];
-                for (String exportedFile : mExportedFiles) {
-                    File file = new File(exportedFile);
-                    FileInputStream fileInputStream = new FileInputStream(file);
-                    zipOutputStream.putNextEntry(new ZipEntry(file.getName()));
-
-                    int length;
-                    while ((length = fileInputStream.read(buffer)) > 0) {
-                        zipOutputStream.write(buffer, 0, length);
-                    }
-                    zipOutputStream.closeEntry();
-                    fileInputStream.close();
-                }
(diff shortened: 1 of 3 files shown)


rivaldi8:
I forgot to run the tests and just found some got broken. I'll look into it.

rivaldi8:
Fixed!

app/src/main/java/org/gnucash/android/export/ExportAsyncTask.java
```diff
@@ -283,25 +283,10 @@ private void moveExportToUri() throws Exporter.ExporterException {
         if (mExportedFiles.size() > 0){
             try {
                 OutputStream outputStream = mContext.getContentResolver().openOutputStream(exportUri);
-                ZipOutputStream zipOutputStream = new ZipOutputStream(outputStream);
```
codinguser: @rivaldi8 While going through the commits to create a CHANGELOG for the next release, I revisited this issue. 
I think we should zip the file which we export (even if it is just a single one), because it saves bandwidth for user sync. Also, some users have really large databases which lead to large XML outputs (always exports the whole database).
 
One way to remedy this would be to make the GncXmlExporter ***always*** exports zipped files (without the .zip extension, only .gnca - .gnca will then be considered a zipped format). Then we do not need to worry about it here (or anywhere else). 

GnuCash desktop can handle zipped XML files for export, as can our app. It will not be directly user-readable, but then again, XML is not really meant to be user-readable. And those who want to read it can just unzip it once. 

What do you think?

rivaldi8: @codinguser Yes, I agree with you, but I think you really mean gzipped (`GZIPOutputStream`, like [here](https://github.com/rivaldi8/gnucash-android/blob/55c8f1817865ada708523e040fb3559747adc0dd/app/src/main/java/org/gnucash/android/export/xml/GncXmlExporter.java#L936-L936)) instead of zipped (`ZipOutputStream`). That's the compression algorithm used in .gnca files, isn't it?

codinguser: @rivaldi8 yes, I meant gzipped, sorry :)

PR #720, opened 2017-09-03
Bugfix 544: Add dialog with extra checkbox for irreversible actions

### LINKED ISSUE #544: Add dialog with extra checkbox for irreversible actions
Actions such as deleting book, deleting all accounts, deleting all transactions

rivaldi8:
I haven't found any other dialog that should have double confirmation, apart from the ones mentioned on the issue, but I might've missed some.

Fixes #544.

### DIFF
--- a/app/src/main/java/org/gnucash/android/ui/settings/BookManagerFragment.java
+++ b/app/src/main/java/org/gnucash/android/ui/settings/BookManagerFragment.java
@@ -38,6 +38,8 @@ import android.view.MenuInflater;
 import android.view.MenuItem;
 import android.view.View;
 import android.view.ViewGroup;
+import android.widget.CheckBox;
+import android.widget.CompoundButton;
 import android.widget.EditText;
 import android.widget.ImageView;
 import android.widget.ListView;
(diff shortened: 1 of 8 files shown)


codinguser:
Thanks for the PR. 
CI build fails: 
org.gnucash.android.test.unit.model.TransactionTest > testCreateAutoBalanceSplit FAILED
    java.lang.NullPointerException at TransactionTest.java:89

rivaldi8:
Thanks, I didn't notice it had failed (sometimes I don't get notifications when the build fails).

Something is wrong with the tests because they pass in my machine. For some reason, it only fails if I change the locale to en_US.UTF-8.

As expected, the problem was introduced in my split-related changes (2353957). I'll investigate what's going on.

codinguser:
ok, that's weird. I just checked it out locally and it builds for me too. 
I retriggered the build on Travis and it failed before. I will trigger a third build again and see. 
But just go ahead and look at the split stuff again

rivaldi8:
I've added a commit that fixes the tests.

You can reproduce the test failure by adding `-Duser.language=en -Duser.country=US` to `org.gradle.jvmargs` in `gradle.properties`. I haven't modified it because it's in `.gitignore` and I don't know if you did so for some reason. [From this discussion](https://discuss.gradle.org/t/best-practice-for-gradle-properties-and-scm/5809/4) it seems it's recommended. It would make running the tests more independent of the environment. In fact, it seems to solve some problems I had with UI tests failing due to locale issues. What do you think?

codinguser:
Thanks @rivaldi8 
Yeah, I added it to .gitignore on purpose. The properties file contains API keys and release build config info which shouldn't be public. 
I assumed that project-level properties file would override my default user-level properties, but this seems not to be the case with gradle - the reverse is true. So I'll consider migrating those variables to my user-level gradle.properties.

Issue #725, opened 2017-09-04
Enhancement :: improvement in input view of transaction


yohji:
I would to suggest two possibile improvement for the input view of a new transaction:

* a new button for pre-fill the entire amount of the current account (useful for empty the account in favor of another account)

* a way to pre-fill the amount of the transation, as the difference between the current amount of the select account and a calculated *delta*, to reach a desired value (useful for account having a *plafond* like a credit card)

yohji:
I realize that the second one can be solved by just using the operation "-", in conjunction with the total amount of the current account (the first one request). I'm studing the code to implement the first one; I will pull it when/if it is ready. Thanks.

codinguser:
Alright. Remember to read our contribution guidelines: https://github.com/codinguser/gnucash-android/blob/master/.github/CONTRIBUTING.md

yohji:
I propose the #732 pull request in response to this issue. Thanks.

codinguser:
See comment on #732

PR #732, opened 2017-09-17
Extended calculator keyboard

yohji:
This pull request is in response to the enhancement issue #725 . It concerns the calculator keyboard:

- introduced a new key on the keyboard indicated by "_T_", that it fills the amount with the total (balance) of the current account.

- fixed the _getBalance_() of _Money_; it loses the value of the incremental adding from iteration.

- fixed the "_C_" in the landscape _res_; it points to the "_back_".

### DIFF
--- a/app/src/main/java/org/gnucash/android/model/Account.java
+++ b/app/src/main/java/org/gnucash/android/model/Account.java
@@ -269,7 +269,7 @@ public class Account extends BaseModel {
     public Money getBalance() {
         Money balance = Money.createZeroInstance(mCommodity.getCurrencyCode());
         for (Transaction transaction : mTransactionsList) {
-            balance.add(transaction.getBalance(getUID()));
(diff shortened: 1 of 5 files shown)


codinguser:
@yohji Thanks for the PR. 
I will evaluate this PR soon and give you feedback.

codinguser:
@yohji I was just going through the PR again. Sorry for the delay in reviewing it. 

A few points:
* The symbol on the button (looks like a "`T`") does not immediately convey what that button does and is thus very opaque. 

* As a consequence of the previous point, I was thinking about how often that button would be used. I think not very often, (indeed you mentioned in #725 that it is useful mostly for emptying the account in favor of another one). This is not something that I anticipate would happen often and therefore I do not think it is worth it to have an extra button in the UI for that. 

I'm sorry, I don't think I will be merging this.  
Thank you for the effort you put into making GnuCash Android better. Keep the feedback coming!

Cheers,

PR #739, opened 2017-10-03
Latest commodities

### LINKED ISSUE #731: Persian currency symbol is doubled
1. Choose IRR as currency symbol

#### Expected behaviour
It should show only one character currency symbol

#### Actual behaviour
It shows two consecutive IRR currency symbols

#### Software specifications
* GnuCash Android version: 2.1.7
* System Android version: 4.2.2
* Device type: Smart phone

Similar bug in the main application and the fix:
[Bug 787479](https://bugzilla.gnome.org/show_bug.cgi?id=787479) - Persian currency symbol is doubled

pnemonic78:
fixes #731

Copied the latest currencies XML file from the GnuCash desktop version with correct IRR symbol.
This requires a re-population of the commodities database table.


app/src/main/java/org/gnucash/android/importer/CommoditiesXmlHandler.java
```diff
@@ -82,6 +82,7 @@ public void startElement(String uri, String localName, String qName, Attributes
 
     @Override
     public void endDocument() throws SAXException {
+        mCommoditiesDbAdapter.deleteAllRecords();
```
codinguser: What is the purpose of this call? Aren't we always importing into a new book anyway? (which means a fresh database)

codinguser:
Thanks for the PR @pnemonic78 
I have a couple of remarks:
1. I would prefer to do a database migration only in `develop` branch as far as is possible. Hotfix db migrations should only be for burning issues. 
2. Please in the future separate refactorings and feature/bugfix pull requests. This PR makes many modifications which are unrelated to the main goal of the PR and that makes it hard to understand where the actual fixes are. 
3. I do not see the new currencies XML in the list of changed files. Where is it saved?
4. I also added a question to the PR itself (you should have gotten a notification).

Cheers,

codinguser:
Thanks for the PR @pnemonic78 
I have a couple of remarks:
1. I would prefer to do a database migration only in `develop` branch as far as is possible. Hotfix db migrations should only be for burning issues. Could you please reissue this against develop?
2. Please in the future separate refactorings and feature/bugfix pull requests. This PR makes many modifications which are unrelated to the main goal of the PR and that makes it hard to understand where the actual fixes are. 
3. I do not see the new currencies XML in the list of changed files. Where is it saved?
4. I also added a question to the PR itself (you should have gotten a notification).

Cheers,

app/src/main/java/org/gnucash/android/importer/CommoditiesXmlHandler.java
pnemonic78: no, an existing book that will have its database "upgraded" means that the table is already populated with commodities, so we need to delete the old ones and populate from the latest xml.

pnemonic78:
1. will re-do in develop
2. they are all related to having the latest and greatest currencies.
3. https://github.com/codinguser/gnucash-android/pull/739/commits/c3754782330cb0f44ce1c93317d16631f3134b82
4. answered above.

pnemonic78:
ready for approval :-)

codinguser:
Just to note that I didn't forget this. I will merge it into develop for the release after 2.4.0 (currently in beta)

PR #742, opened 2017-10-30
Android Studio 3 compatibility

MisterY:
A few things to make the project build with Android Studio 3:

-  updating the build tools
- adding flavor dimensions
- handling output file name
- updating Gradle plugin & Gradle

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -21,7 +21,7 @@ static def gitSha() {
 
 android {
     compileSdkVersion 26
-    buildToolsVersion '26.0.1'
+    buildToolsVersion '27.0.0'
     defaultConfig {
         applicationId "org.gnucash.android"
         testApplicationId 'org.gnucash.android.test'
@@ -49,9 +49,9 @@ android {
     }
 
     applicationVariants.all { variant ->
-        variant.outputs.each { output ->
-            output.outputFile = new File(
-                    output.outputFile.parent, "GnucashAndroid_v${variant.versionName}.apk")
+        variant.outputs.all { output ->
+            //output.outputFile = new File(output.outputFile.parent, "GnucashAndroid_v${variant.versionName}.apk")
+            outputFileName = "GnucashAndroid_v${variant.versionName}.apk"
         }
     }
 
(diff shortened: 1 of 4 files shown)


MisterY:
This is needed in order to proceed with other additions but I decided to send it separately. It is not directly related to other changes I'd like to make.

app/build.gradle
```diff
@@ -49,9 +49,9 @@ android {
     }
 
     applicationVariants.all { variant ->
-        variant.outputs.each { output ->
-            output.outputFile = new File(
-                    output.outputFile.parent, "GnucashAndroid_v${variant.versionName}.apk")
+        variant.outputs.all { output ->
+            //output.outputFile = new File(output.outputFile.parent, "GnucashAndroid_v${variant.versionName}.apk")
```
rivaldi8: Unless there's some good reason to leave the old line commented, please, remove it. Git already takes care of remembering it.

app/build.gradle
```diff
@@ -104,6 +106,8 @@ android {
             resValue "string", "dropbox_app_key", "dhjh8ke9wf05948"
             resValue "string", "dropbox_app_secret", "h2t9fphj3nr4wkw"
             resValue "string", "manifest_dropbox_app_key", "db-dhjh8ke9wf05948"
+
+            dimension "stability"
```
rivaldi8: When there's just one dimension, it's assigned by default to all flavors, you don't need to specify it. See [here](https://google.github.io/android-gradle-dsl/current/com.android.build.gradle.AppExtension.html#com.android.build.gradle.AppExtension:flavorDimensions(java.lang.String[])).

app/build.gradle
```diff
@@ -120,6 +124,8 @@ android {
                 resValue "string", "dropbox_app_secret", "h2t9fphj3nr4wkw"
                 resValue "string", "manifest_dropbox_app_key", "db-dhjh8ke9wf05948"
             }
+
+            dimension "stability"
```
rivaldi8: Same here. No need to specify it.

app/build.gradle
```diff
@@ -135,6 +141,8 @@ android {
...
```
rivaldi8: Same here.

MisterY:
Thanks for the feedback! I've applied the changes. Looking forward to making some functional improvements/additions in the near future.

PR #755, opened 2017-12-30
Recover from empty books database

rivaldi8:
The fix implemented in c27f763 didn't solve the issue. It now fails because the books database is empty. This pull request will hopefully solve the problem, unless there's been real data loss.

Fixes http://crashes.to/s/d7c34870c8d


codinguser:
Thanks @rivaldi8 
The fix looks good and comprehensive to me. 
I hope I can eventually figure out at what point the issue is triggered in the first place which causes no book to be active.

Issue #756, opened 2018-01-04
Export book in CSV format


codinguser:
GnuCash desktop will include a new CSV importer which can handle splits and prices etc. 
Would be good if the app can support this format and maybe eventually switch to CSV export as the default. 

Template for the CSV format would be that which GnuCash desktop itself currently exports. (Up till v2.6, it was [not possible for GnuCash desktop to consume it's own CSV exports](http://gnucash.1415818.n4.nabble.com/CSV-Import-tp4690603p4690639.html). But that should change from v3.0)

PR #758, opened 2018-01-10
CSV Export

sononio:
This fork adds csv export variant with separator choosing. 

There are two variants of export:

- Acconts info
- Transactions info

There are three variants of separator:
 - "," - comma
- ":" - colon
- ";" - semicolon

### DIFF
--- a/app/src/main/java/org/gnucash/android/export/csv/CsvTransactionsExporter.java
+++ b/app/src/main/java/org/gnucash/android/export/csv/CsvTransactionsExporter.java
@@ -1,0 +1,237 @@
+/*
+ * Copyright (c) 2018 Semyannikov Gleb <nightdevgame@gmail.com>
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.gnucash.android.export.csv;
+
+import org.gnucash.android.export.xml.*;
+
+import android.database.Cursor;
+import android.database.sqlite.SQLiteDatabase;
+import android.net.Uri;
+import android.util.Log;
+
+import com.crashlytics.android.Crashlytics;
+
+import org.gnucash.android.app.GnuCashApplication;
+import org.gnucash.android.db.DatabaseSchema;
+import org.gnucash.android.db.adapter.BooksDbAdapter;
+import org.gnucash.android.db.adapter.CommoditiesDbAdapter;
+import org.gnucash.android.db.adapter.RecurrenceDbAdapter;
+import org.gnucash.android.db.adapter.TransactionsDbAdapter;
+import org.gnucash.android.export.ExportFormat;
+import org.gnucash.android.export.ExportParams;
+import org.gnucash.android.export.Exporter;
+import org.gnucash.android.model.Account;
+import org.gnucash.android.model.AccountType;
+import org.gnucash.android.model.BaseModel;
+import org.gnucash.android.model.Book;
+import org.gnucash.android.model.Budget;
+import org.gnucash.android.model.BudgetAmount;
+import org.gnucash.android.model.Commodity;
+import org.gnucash.android.model.Money;
+import org.gnucash.android.model.PeriodType;
+import org.gnucash.android.model.Price;
+import org.gnucash.android.model.Recurrence;
+import org.gnucash.android.model.ScheduledAction;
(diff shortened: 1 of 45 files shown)


app/src/main/java/org/gnucash/android/export/csv/CsvTransactionsExporter.java
```diff
@@ -0,0 +1,177 @@
...
+            names.add("Action");
+            names.add("Reconcile");
+            names.add("To With Sym");
+            names.add("From With Sym");
+            names.add("To Num.");
+            names.add("From Num.");
+            names.add("To Rate/Price");
+            names.add("From Rate/Price");
+
+            List<Transaction> transactions = mTransactionsDbAdapter.getAllTransactions();
```
codinguser: The number of transactions in the database can be quite big for very large books over time. It is not efficient to load everything into memory. It is better to use a cursor to iterate over the database and retrieve the entries you need on-demand. 
I know just getting all records is easier at first, but it will become technical debt in the future

app/src/main/java/org/gnucash/android/export/csv/CsvTransactionsExporter.java
```diff
@@ -0,0 +1,177 @@
...
+
+            for(int i = 0; i < names.size(); i++) {
+                writer.write(names.get(i) + separator);
+            }
+            writer.write("\n");
+            for(int i = 0; i < transactions.size(); i++) {
+                Transaction transaction = transactions.get(i);
+                List<Split> splits = transaction.getSplits();
+                for (int j = 0; j < splits.size()/2; j++) {
```
codinguser: This loops depends on the ordering of the splits in the list. Even if it might be the case at the moment, I would prefer it we did not depend on it. The logic here would change anyway when switching to cursor-based iteration. When modifying it, please do not rely on the ordering of the splits.

app/src/main/java/org/gnucash/android/ui/export/ExportFormFragment.java
```diff
@@ -197,12 +207,37 @@ private void onRadioButtonClicked(View view){
...
                 }
 				mExportDateLayout.setVisibility(View.VISIBLE);
+				mExportSeparatorLayout.setVisibility(View.GONE);
 				break;
 
 			case R.id.radio_xml_format:
 				mExportFormat = ExportFormat.XML;
 				mExportWarningTextView.setText(R.string.export_warning_xml);
 				mExportDateLayout.setVisibility(View.GONE);
+				mExportSeparatorLayout.setVisibility(View.GONE);
```
codinguser: The mExportSeparatorLayout does not seem to include the `Separator` label. If you open the app and switch to CSV and then back to XML, the label `Separator` is still visible.

app/src/main/res/values-af-rZA/strings.xml
```diff
@@ -179,6 +179,8 @@
     <item>QIF</item>
     <item>OFX</item>
     <item>XML</item>
+    <item>CSV-accounts</item>
```
codinguser: The translation is handled by the CrowdIn platform. No need to do it in the translation files. 
I should add this to the development guidelines.

app/src/main/java/org/gnucash/android/export/csv/CsvTransactionsExporter.java
sononio: I'll try to fix it in 5-10 days. Now i have trouble because i have no big db. Can you help me and give some example of big enough book in xml? I need it to test export speed and compatibility with gnucash desktop

app/src/main/java/org/gnucash/android/export/csv/CsvTransactionsExporter.java
sononio: I'll try to fix it in 5-10 days

app/src/main/res/values-af-rZA/strings.xml
sononio: I'll try to fix it in 5-10 days

app/src/main/java/org/gnucash/android/export/csv/CsvTransactionsExporter.java
codinguser: Hmm, I do not have any big db. I've been thinking to write some modules in the test framework which generates tons of transactions. But I never got around to it.

No matter, I think you can just switch it to cursor iteration without having to actually have a big book to test it on. Users who have them will appreciate it.

codinguser:
I'm reviewing the PR.

codinguser:
Thanks for the updates. Sorry it took a while to review it. 
I think you missed one of my inline code comments. 
* The `Separator` label (not the actual options, just the label) is still visible in the UI even when the export format is not CSV. Could you make it show up only for CSV exports? 

Also, there were two questions from my first review which you didn't respond to yet

> Also a few questions:

> * Are these two CSV formats supported by GnuCash desktop? As far as I understand, there is support in the newer (unstable) and future 3.0 version for importing CSV. Please ensure that the CSV we generate can be consumed by the desktop version. See #756
> * Is there any CSV format which supports both accounts and transactions in one file?
> 

Thanks again!

codinguser:
@sononio I made the changes to the separator label and merged. 
Cheers,

Issue #777, opened 2018-05-01
Compatibility with Desktop V 3


pduvallet:
Import XML data from desktop V3

Currency are not recognized

#### Software specifications
Desktop V 3.1
* GnuCash Android version: 2.3
* System Android version: 7.0
* Device type: Samsung Galaxy S7

rafaelrmrr:
Hi,
I'm with the same issue. Currency are not recognized and the account's balance, is another problem. Shows only XXX0,000000

GnuCash Desktop
Versão: 3.1
Build ID: 3.0-118-gd2ef5fd0f+ (2018-04-28)

Smartphone Moto Z2 Play

The account's balance, is another problem, shows only XXX0,000000

PR #779, opened 2018-05-08
Added ability to load GNUCash v. 3.0 xml

xzfantom:
#777 Now it's possible to load xml from desktop version 3.0

### DIFF
--- a/app/src/main/java/org/gnucash/android/importer/GncXmlHandler.java
+++ b/app/src/main/java/org/gnucash/android/importer/GncXmlHandler.java
@@ -432,7 +432,7 @@ public class GncXmlHandler extends DefaultHandler {
                 mAccount.setHidden(accountType == AccountType.ROOT); //flag root account as hidden
                 break;
             case GncXmlHelper.TAG_COMMODITY_SPACE:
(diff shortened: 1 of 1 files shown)


codinguser:
Thanks @xzfantom

businessBoris:
Oh man! I've just loaded gnucash (Linux desktop and mobile) again to try after many years hiatus.
This is a present from the gods.
Thank you very very much.

UPDATE: I was so excited that I was minded to reset SyncThing for the new setup now! Forgot about a little thing like the actual release of this update to 'production'! Doh!

codinguser:
@businessBoris you could join the beta program if you would like to get your hands on it fast. It will hit there first

codinguser:
Alright, no worries then. It'll come when it's fully baked.

businessBoris:
Where do I find the find the actual xml on Android? I think it's hidden from me on Android 7.0
Happy to post to the wiki if that's a thing you want to do.

xzfantom:
Gnucash Android don't use xml to store data, it uses internal SQLite database. Xml is used for import only.

businessBoris:
Thank you.

PR #782, opened 2018-05-25
Misuse of multiple SQL statements in execSQL()

lordpyre:
Hello,
I've spotted misuses of multiple SQL statements in some execSQL() calls. In fact, the triggers do not get created by SQLite like this. Please consider reviewing and fixing it.
Best,
Csaba


codinguser:
Hi, thanks for the PR. 
We do not accept merges on `master`. I do not think there is an issue in the `develop` branch.

PR #791, opened 2018-06-11
text margins for RTL text in transaction details screen

### LINKED ISSUE #790: Transaction description margin
1. Transaction screen
2. Transaction description in RTL language, e.g. Hebrew

#### Expected behaviour
Margin on "right" edge of screen

#### Actual behaviour
No margin

image_1528407021661

pnemonic78:
fixes #790


pnemonic78:
Don't confuse RTL content with RTL layouts.
`android:layout_marginStart` is "left" to LTR layouts, but "right" for RTL layouts. For example, the "x" button appears on top-left in LTR layout, but will be in top-right in RTL layout => title margin should be on left => margin should be "end".

Cannot just add `android:supportsRTL` without checking that **all** the screens are still OK.

PR #802, opened 2018-11-01
translate to farsi

mohsensaleh2010:
translate app menu to Iranian Persian language


codinguser:
Thanks very much for the PR. I think you already translated on the CrowdIn page. I will pull the translations from there. 
Also, for future ref, all PRs should be against develop, not master.

PR #816, opened 2019-02-15
Develop (Fixed overlapping of time text in Export Transaction Activity)

Anil-Praharaj21:
Issue pic (At Time 3:56:15 PM was hiding)
screenshot_20190214-180857

Fixed this issue
screenshot_20190214-184639

### DIFF
--- a/app/src/main/java/org/gnucash/android/ui/account/AccountsActivity.java
+++ b/app/src/main/java/org/gnucash/android/ui/account/AccountsActivity.java
@@ -227,6 +227,7 @@ public class AccountsActivity extends BaseDrawerActivity implements OnAccountCli
 
         init();
 
+        getSupportActionBar().setSubtitle(BooksDbAdapter.getInstance().getActiveBookDisplayName());
         TabLayout tabLayout = (TabLayout) findViewById(R.id.tab_layout);
         tabLayout.addTab(tabLayout.newTab().setText(R.string.title_recent_accounts));
         tabLayout.addTab(tabLayout.newTab().setText(R.string.title_all_accounts));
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/gnucash/android/ui/transaction/TransactionFormFragment.java
```diff
@@ -139,7 +140,7 @@
 	/**
 	 * Formats a {@link Date} object to time string of format HH:mm e.g. 15:25
 	 */
-	public final static DateFormat TIME_FORMATTER = DateFormat.getTimeInstance();
+	public final static DateFormat TIME_FORMATTER = new SimpleDateFormat("hh:mm a");
```
codinguser: This does not take into account time localization. Not all users have an AM/PM time display format

app/src/main/java/org/gnucash/android/ui/account/AccountsActivity.java
```diff
@@ -227,6 +227,7 @@ public void onCreate(Bundle savedInstanceState) {
 
         init();
 
+        getSupportActionBar().setSubtitle(BooksDbAdapter.getInstance().getActiveBookDisplayName());
```
codinguser: Might be better to create a separate pull request for this. 
Also, if possible, style the subtitle smaller so that the main title does not feel pushed up too much. 

How about the other activities in the app. Might be good to think a bit about how the book name could be integrated into them as well.

codinguser:
Thanks for the PR. But it cannot be merged in its current form. 
I left some comments in the code review on what could be improved.

Issue #809, opened 2019-01-24
Export to csv - App breaks down


rainerfeldbruegge:
#### Steps to reproduce the behaviour
1. Pick Menu Export, 
 export to file/owncloud/mail (same behaviour for all)
 format csv
time from 01.01.2019
delete transactions after export: no
repeat: none

Tip "Exportieren"

#### Expected behaviour
A file is created in csv-format

#### Actual behaviour
gnu-cash breaks down, Application halted
no file created.
whenn i pick export to "speichern unter" there is a file created, but empty.

#### Software specifications
* GnuCash Android version: 2.4.0
* System Android version: 6.0.1
* Device type: FP 2 (FairPhone)

jainharsh96:
I am working on this issue.

jainharsh96:
I resolved the issue

PR #817, opened 2019-02-15
resolve issue  when save file #809

jainharsh96:
Now the app does not crash.
And it asks to select destination person to send file


codinguser:
Fixed in f69a6090cbbd7e0c96371a36807535f7f28dfa79

PR #826, opened 2019-03-01
Initial release of croatian accounts

milotype:
Initial release of the common croatian accounts. These are the same as the ones I did for the desktop version of GnuCash, plus specifying the account colors defined in the Android version.


codinguser:
Thanks

PR #838, opened 2019-06-08
Updated version of libraries

herou:
added build-tools to grandlew


codinguser:
Thanks @herou . 
I don't think this should be a hotfix. It should go into develop. 
We will be updating the library versions there anyway to the newest ones for the next release.

PR #845, opened 2019-10-06
Fix slow CSV exports

Pikrass:
CSV exporting was taking tens of minutes on my phone, for only around 120 exported transactions, and a database containing 2 years worth of personal finances.
This PR fixes it. Technical details are in commit messages, but basically: the code fetched all transactions from one account for every single exported split. That meant easily tens of thousands or even hundreds of thousands requests for one export.


codinguser:
Thanks!

PR #854, opened 2019-12-26
Removing time zone in TIME_FORMATTER to comply with desktop gnucash's…

DunLug:
Hello everyone,
I had some issue when importing my account file generated with gnucash on windows, version 3.5 (build 3.5+ (2019-03-30). After doing a debug, it seems that the date format in my xml file doesn't include a timezone while the TIME_FORMATTER value looks for it.
I removed the timezone to make importation working, but I don't know if some other versions would include timezone in the xml file.

### DIFF
--- a/app/src/main/java/org/gnucash/android/export/xml/GncXmlHelper.java
+++ b/app/src/main/java/org/gnucash/android/export/xml/GncXmlHelper.java
@@ -147,7 +147,7 @@ public abstract class GncXmlHelper {
 
     public static final String RECURRENCE_VERSION           = "1.0.0";
     public static final String BOOK_VERSION                 = "2.0.0";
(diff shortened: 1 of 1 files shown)


codinguser:
Hi @DunLug The latest version of Gnucash (4.2) still spits out dates like "2020-01-01 21:29:35 +0000" when creating XML files.
So I'm not sure if that was an issue with 3.5 which is now changed or not. 
Can you confirm if this is still an issue?

DunLug:
With gnucash  4.2  the format is 
<trn:date-posted>
    <ts:date>2020-12-02 10:59:00 +0000</ts:date>
  </trn:date-posted>

With gnucash 3.5 it was 
<trn:date-posted>
    <ts:date>2015-11-03 10:59:00</ts:date>
</trn:date-posted>

I checked the import  on android using the 2.4.0 version on google play and it works fine.
I close the ticket while this issue was resolved with newest versions

PR #860, opened 2020-01-05
Use "License" everywhere for consistency

jeffwidman:
While "Licence" is an allowed spelling, the rest of the paragraph uses "License" so switch to that for consistency.


jeffwidman:
Nudge

jeffwidman:
nudge @codinguser

codinguser:
Thanks @jeffwidman. I merged this before realizing that it was going to master not develop. 
We only merge into `develop` branch. But it's in now and I will just leave it as-is merge master back into develop

Issue #861, opened 2020-01-07
Use ₽ sign to designate Russian ruble, instead of "руб"


lompy:
Issue from the [forum thread](https://gnucash.uservoice.com/forums/320493-gnucash-android/suggestions/32571610-use-sign-to-designate-russian-ruble-instead-of)

> It is incorrect to use "руб" as a prefix for the amount of money. One can either write "₽500" or "500 руб." (with white space and dot). For this application I recommend to use the first one.

lompy:
Also the last letter of the "руб" tends to be interpreted as the first digit "6" of the number: "руб105.91". Highly unreadable.

codinguser:
develop branch already has the symbol ₽

codinguser:
Will be available in next release

PR #862, opened 2020-01-07
Use ₽ sign to designate Russian ruble, instead of "руб"

lompy:
https://github.com/codinguser/gnucash-android/issues/861


codinguser:
Thanks @lompy 
The currencies file is taken from the main GnuCash desktop project. I have updated the sources in 3ffeb5e68dedb90807f4e86b7222a09c8555c88b
Taken from: https://github.com/Gnucash/gnucash/blob/maint/libgnucash/engine/iso-4217-currencies.xml

codinguser:
Actually, the symbol is already updated in the develop branch to ₽

PR #865, opened 2020-01-09
Fixes #856 - Fix the Bad Preference Title at first opening and Naviga…

### LINKED ISSUE #856: Bad Preference Title at first opening
1. Open Preferences

#### Expected behaviour
Title should be "Preferences"

#### Actual behaviour
Title is "General Preferences", which is the title of the ActionBar if you choose the "General" list item

If you click "General" list item, it opens with "General Preferences"
If you slide to the right, you get back the Preference List, but this time, the title is the expected one : "Preferences"

#### Software specifications
    GnuCash Android version: 2.4.0
    System Android version: 9.0
    Device type: Samsung Galaxy Note 8

JeanGarf:
…tion View "Preferences" label

When you open the Main menu, the item "settings" open a "General Preferences" Activity but it should be called "Settings", not "General Preferences". "General Preferences" is the right title if you click on the "General" item. The problem was the call to GeneralPreferenceFragment() line 96 in PreferenceActivity which override the "Settings" title

All the titles are "Something Preferences" (General Preferences, Account Preferences, Transaction Preferences...), therefore the item in the main menu should be "Preferences" instead of "Settings". Same in the ActionBar where "Settings" has been replaced by "Preferences".

I changed these two things in this Pull Request.

I also changed in French version. I've seen in "Contributing Guidelines" that translation should be made using CrowdIn, but the link is broken and I don't know how to use CrowdIn.

### DIFF
--- a/app/src/main/res/values-fr/strings.xml
+++ b/app/src/main/res/values-fr/strings.xml
@@ -54,7 +54,7 @@
   <string name="btn_export">Exporter</string>
   <string name="option_delete_after_export">Effacer les transactions après export</string>
   <string name="hint_delete_after_export">Toutes les transactions exportées seront supprimées après l\'export</string>
-    <string name="title_settings">Préférences</string>
+    <string name="title_settings">Préférences / Maintenance</string>
   <string-array name="export_destinations">
     <item>Enregistrer sous &#8230;</item>
     <item>Dropbox</item>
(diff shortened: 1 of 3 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -95,7 +95,7 @@
 	<string name="title_transaction_preferences">Transaction Preferences</string>
 	<string name="title_account_preferences">Account Preferences</string>
 	<string name="title_default_transaction_type">Default Transaction Type</string>
-	<string name="summary_default_transaction_type">The type of transaction to use by default, CREDIT or DEBIT</string>
+	<string name="summary_default_transaction_type">CREDIT</string>
```
codinguser, line 98: This is the default summary which is shown when a default transaction type has not been set. 
The goal is to explain the setting. Once set, the currently set value will be displayed

app/src/main/res/values-fr/strings.xml
```diff
@@ -54,7 +54,7 @@
   <string name="btn_export">Exporter</string>
```
codinguser, line 54: Unfortunately, all changes in this file will be overriden after importing from CrowdIn, please do the transalations there

codinguser:
Thanks @JeanGarf 
I would prefer not to merge this because the translations would just get overridden. 
Please do translations in CrowdIn. Or you could open a PR just with the part which fixes the display of "General Preferences" instead of "Settings" when the settings drawer is opened

Issue #855, opened 2020-01-04
Press Back Button twice to quit GnuCashAndroid


JeanGarf:
#### Steps to reproduce the behaviour
1. On Accout Activity, clicking Back button once quit GnuCashAndroid

#### Expected behaviour
1. It would be nice to have the option to click Back button twice to avoid quitting unexpectedly
2. A General Preference allow to enable/disable this function

#### Software specifications
* GnuCash Android version: 2.4.0
* System Android version: 9.0
* Device type: Samsung Galaxy Note 8

JeanGarf:
I'have corrected this issue in 7b2b45fe9fe66086ee41b8c1dfbab4f106eb83f0, but I don't know yet how to give it to you

PR #867, opened 2020-01-09
Fixes #859 - Jump to other Account Transaction List

### LINKED ISSUE #859: Jump to other Account Transaction List
1. Open the transaction list for an Account
2. Click on the grey "other account name" in the middle of the list item
3. Click on the Transaction title or in the white zone

#### Expected behaviour
2. The "other account" transaction list opens
3. The Transaction Detail Activity opens

#### Actual behaviour
2. The Transaction Detail Activity opens
3. The Transaction Detail Activity opens

#### Software specifications
* GnuCash Android version: 2.4.0
* System Android version: 9.0
* Device type: Samsung Galaxy Note 8

JeanGarf:
This Fix provides the "Jump" function of GnuCash (on Windows).

When you'are on an account transaction list :
1) Il you click on the transaction text title or elsewhere in the white zone of the item, it displays the details of the transaction as before.

2) But now, if you click on the other account (secondery text) of a transaction, the other account transaction list appears.
Pressing the back button unpops and jump back to the previous account transaction list

[Video demo #855 to #859.zip](https://github.com/codinguser/gnucash-android/files/4043120/Video.demo.855.to.859.zip)

### DIFF
--- a/app/src/main/java/org/gnucash/android/ui/transaction/TransactionsListFragment.java
+++ b/app/src/main/java/org/gnucash/android/ui/transaction/TransactionsListFragment.java
@@ -169,12 +169,20 @@ public class TransactionsListFragment extends Fragment implements
 		refresh();
 	}
 
-	public void onListItemClick(long id) {
-		Intent intent = new Intent(getActivity(), TransactionDetailActivity.class);
-		intent.putExtra(UxArgument.SELECTED_TRANSACTION_UID, mTransactionsDbAdapter.getUID(id));
+    /**
+     * Called when user clicks on a transaction list item
+     *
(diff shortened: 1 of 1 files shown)


codinguser:
Thanks @JeanGarf 
I have some thoughts. I think that the transaction text in the list view is a really small target to hit. 
Also, when the transfer account name is long, it stretches out over the card and you might end up jumping to the account when you just meant to open it.

I would suggest instead to implement the jump from inside the transaction detail view. 
That way, it will also work for multiple splits where from the transaction view, you can jump to any of the split accounts.
It might be necessary to increase the height of the account view in the split list to make it an easier target to hit.

PR #900, opened 2021-06-02
Fix #851 - Create Sub-accounts screen should have default currency same as parent, not app-default currency

### LINKED ISSUE #851: Create Sub-accounts screen should have default currency same as parent, not app-default currency
Create Sub-accounts screen should have default currency same as parent, not app-default currency. This makes it convenient to create a parallel currency tree by not having to select the parent's currency again and again.

1. Create an account of a currency (say GBP) which is different from app default currency (say USD).
2. Tap on "+" to create a sub-account.

#### Expected behaviour
The currency selected by default should be GBP

#### Actual behaviour
The currency selected by default is USD

#### Software specifications
* GnuCash Android version: 2.4.0
* System Android version: 9
* Device type: OnePlus 6T

RasheeqZaman:
This is more like a feature rather than a fix. When creating a sub-account, It'll be helpful for the users if the default currency was the same as Parent's currency.

### DIFF
--- a/app/src/main/java/org/gnucash/android/ui/account/AccountFormFragment.java
+++ b/app/src/main/java/org/gnucash/android/ui/account/AccountFormFragment.java
@@ -430,6 +430,7 @@ public class AccountFormFragment extends Fragment {
             setAccountTypeSelection(parentAccountType);
             loadParentAccountList(parentAccountType);
             setParentAccountSelection(mAccountsDbAdapter.getID(mParentAccountUID));
+            setSelectedCurrency(mAccountsDbAdapter.getCurrencyCode(mParentAccountUID));
         }
 
     }


kmdasstech:
Hello Rasheeq,
I am new to GitHub. Can you please let me know how to download the app with all latest changes?

RasheeqZaman:
Hi kmdasstech, 
I am new to working on open source projects too. So I am not sure if my way is correct. But I'll tell you what I did.

- Fork the project. It will create a copy of the project in your github account.
- In your forked project, Click on the "code" button. You will see a link of the project in there. Copy that link.
- Open Android Studio. (If you have any existing project already open then close that project)
- Click on "Get from Version Control".
- Paste that link in the "URL" textfield.
- Click "Clone".

That should do it. If you have any problem in any step you can search google about that step. I am sure you will easily find some useful youtube tutorial about that. Or you could also comment here about the problem. I would be happy to help.

kmdasstech:
Hello Rasheeq, 
Thank you for your instruction. I will try to do so.

